PR #21, opened 2021-11-30
Add OpenBuildings Part Class

wouter-rombouts-bentley:
OpenBuildings products refer to "Part" instances from every element. "Parts" define resymbolization behavior and are stored in the workspace. By Persisting the Parts in the itwin we can monitor changes and update resymbolization when needed.


rschili (dismissed):
Looks good to me.

ColinKerr (approved):

PR #35, opened 2022-01-31
Commit in Jonas' behalf, capturing latest changes regarding v2.0.0.

diegoalexdiaz:


Domains/Grids/Grids.ecschema.xml
```diff
@@ -3,59 +3,15 @@
...
-        <BaseClass>bis:ElementDrivesElement</BaseClass>
-        <Source multiplicity="(1..1)" roleLabel="drives" polymorphic="true">
-            <Class class="Grid"/>
-        </Source>
-        <Target multiplicity="(0..*)" roleLabel="is driven by" polymorphic="true">
-            <Class class="GridSurface"/>
-        </Target>
-    </ECRelationshipClass>
-
     <ECRelationshipClass typeName="GridAxisContainsGridSurfaces" modifier="None" strength="embedding" description="maps axis to grouped surfaces">
```
diegoalexdiaz, line 15: I believe your question on email regarding whether an InformationRecord could own a SpatialLocation element corresponds to this relationship, correct? In BIS terms, if this case is truly a "parent owns children" situation, the parent (GridAxis) will be forced to be a SpatialLocationElement. 

The current "contained in" role seems strange too, considering it is defined between an InformationRecord and a SpatialLocation element.

I haven't wrapped my head around the concept of a GridAxis and why it is introduced as an InformationRecord. I'll need to catch up on that design decision in order to have a stronger opinion about what BIS relationship is more appropriate in this case.

PR #83, opened 2022-04-14
Introduced generating schema jsons and publish them as an artifact

hl662:
Produce a build artifact for bisschema jsons, allowing it to be consumed in internal release pipelines


tools/packages/packageSchema.yaml
```diff
@@ -56,3 +56,52 @@ stages:
...
+      - powershell: |
+          $npmrcDir = "$(Build.Repository.LocalPath)"
+          
+          if(!(Test-Path -Path $npmrcDir)){
+              mkdir $npmrcDir
+          }
+          
+          $npmrcFile = $npmrcDir + "/.npmrc"
+          
+          Out-File -FilePath $npmrcFile -Encoding ascii -InputObject "@bentley:registry=https://registry.npmjs.org/"
```
ColinKerr, line 93: Do we need this?  We're just pulling from public npm

ColinKerr (approved):

tools/packages/packageSchema.yaml
hl662: removed the need for seeding npm rc in previous commit

PR #87, opened 2022-04-28
Introduction of Ports into DistributionSystems schema

diegoalexdiaz:


JValiunas (dismissed):
Just as i remember :)

Domains/1-Common/DistributionSystems/DistributionSystems.ecschema.xml
```diff
@@ -62,4 +62,83 @@
             </IsMixin>
         </ECCustomAttributes>
     </ECEntityClass>
+
+    <ECEnumeration typeName="DirectionType" backingTypeName="int" isStrict="true">
```
karolis-zukauskas, line 66: Perhaps just `Direction` instead of `DirectionType`?

Domains/1-Common/DistributionSystems/DistributionSystems.ecschema.xml
```diff
@@ -62,4 +62,83 @@
...
+    <ECEntityClass typeName="DistributionPort" modifier="Abstract" displayLabel="Distribution Port" description="An inlet or outlet of a distribution element through which a particular substance may flow.">
+        <BaseClass>bis:SpatialLocationElement</BaseClass>
+        <ECProperty propertyName="Description" typeName="string" displayLabel="Description" description="A human-readable description of the Distribution Port"/>
+        <ECProperty propertyName="Direction" typeName="DirectionType" displayLabel="Direction" description="Direction of the port: e.g. Input, Output, Free Directional, Bi-Directional."/>
+        <ECProperty propertyName="LocalOrigin" typeName="point3d" description="The local placement origin of this DistributionPort in relation to its parent.">
+            <ECCustomAttributes>
+                <HiddenProperty xmlns="CoreCustomAttributes.01.00.03"/>
+            </ECCustomAttributes>
+        </ECProperty>
+        <ECProperty propertyName="LocalYaw" typeName="double" description="The local Yaw angle (in degrees) of the orientation of this DistributionPort in relation to its parent.">
```
karolis-zukauskas, line 83: Isn't it against our guidelines to store values not in SI? Shouldn't all angles be stored in radians?

Domains/1-Common/DistributionSystems/DistributionSystems.ecschema.xml
```diff
@@ -62,4 +62,83 @@
...
+        <ECProperty propertyName="LocalRoll" typeName="double" description="The local Roll angle (in degrees) of the orientation of this DistributionPort in relation to its parent.">
+            <ECCustomAttributes>
+                <HiddenProperty xmlns="CoreCustomAttributes.01.00.03"/>
+            </ECCustomAttributes>
+        </ECProperty>
+    </ECEntityClass>
+    
+    <ECRelationshipClass typeName="DistributionElementOwnsDistributionPorts" strength="embedding" modifier="None" description="A relationship that relates child distribution ports to parent distribution element.">
+        <BaseClass>bis:ElementOwnsChildElements</BaseClass>
+        <Source multiplicity="(0..1)" roleLabel="owns" polymorphic="true">
```
karolis-zukauskas, line 102: Shouldn't Source multiplicity be (1..1)?

Domains/1-Common/DistributionSystems/DistributionSystems.ecschema.xml
karolis-zukauskas: Or `DistributionPortDirection`, `PortDirection`?

Domains/1-Common/DistributionSystems/DistributionSystems.ecschema.xml
```diff
@@ -62,4 +62,83 @@
...
+        </Target>
+    </ECRelationshipClass>
+    
+    <ECRelationshipClass typeName="PortConnectionIsRealizedByPhysicalElements" strength="referencing" modifier="None" description="Indicates that the Port Connection is realized by some Physical Element(s).">
+        <BaseClass>bis:ElementRefersToElements</BaseClass>
+        <Source multiplicity="(0..1)" roleLabel="is realized by" polymorphic="true">
+            <Class class="PortConnection"/>
+        </Source>
+        <Target multiplicity="(0..*)" roleLabel="realizes" polymorphic="true">
+            <Class class="bis:PhysicalElement"/>
```
laurynas111, line 140: Would it make sense to change Target constraint to GeoemtricElement3d? To enable use by non-PhysicalElements, e.g. Analytical, or perhaps even Functional?

JValiunas: we should be able to inject a baseclass relationship later. 
so we would still have this particular relationship with constraints as defined here, 

Furthermore, rather than this single relationship, in which case we would also need a base class for DistributionPort and PortConnection, in some now schema.. Maybe we could call it ElementConnectivity. which brings us to a whole new discussion, so best defer it for later.

do you have a sample use case in mind where we would have analytical elements connected by ports?

Domains/1-Common/DistributionSystems/DistributionSystems.ecschema.xml
JValiunas: we seem to have established a standard of suffixing our enums with "Type":
SectionType / AutoHandledPropertyStatementType /CustomHandledPropertyStatementType 

i suggest we follow it here.

Domains/1-Common/DistributionSystems/DistributionSystems.ecschema.xml
JValiunas: can't recall why is this in angles, but I agree, we should also be able to set angle koq here.

Domains/1-Common/DistributionSystems/DistributionSystems.ecschema.xml
JValiunas: in the linked PR we seem to have settled on "Value" suffix? is that the new standard proposed?

my point was about us having some history publishing enums in biscore schema, so I suggest we either suffix it with Type or Value, otherwise we're making a mess...

caseymullen: Grepping all schemas, the only place that seems to have established a clear standard is CIF, where they all end with "_Enum". In BisCore, 3 of 4 end in Type, but that is because they are semantically trying to convey the name of a type or kind of something... not just a potential value and the ECProperty they are paired with ends in "Type". Imagine an enum of standard allowable diameters... it is not exactly a type.
The 4th in BisCore ends in "Role".

I think what BisCore and most authors have done is just name the enum to match the property where it is used.

I agree with Karolis that reserving "Type" to refer to TypeDefinitionElement is more consistent, so if we were going to put an arbitrary suffix on all Enums, that "Enum" would be more appropriate.

So, I think the most common pattern may currently be 
<ECProperty propertyName="Foo" typeName="Foo"
It is subtle, but the way that you recognize it as an enum is that the typeName is anything other than a standard datatype name.

Domains/1-Common/DistributionSystems/DistributionSystems.ecschema.xml
JValiunas: in the discussion, Diego pointed out the same properties in BisCore are also stored in degrees, so we decided to follow here.

Domains/1-Common/DistributionSystems/DistributionSystems.ecschema.xml
JValiunas: we decided for no suffix on enums in our discussion. renamed to Portdirection

JValiunas (approved):

PR #92, opened 2022-05-25
Add new SchemaUpgradeCustomAttributes standard schema

AkshayN5:
Add config attributes standard schema to allow some changes to the schema which otherwise are not allowed.


Standard/Released/ConfigAttributes.01.00.00.ecschema.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- ==================================================================================
+|  * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+|  * See LICENSE.md in the project root for license terms and full copyright notice.
+======================================================================================= -->
+<ECSchema schemaName="ConfigAttributes" alias="ConfigA" version="01.00.00" description="Config attributes to allow some changes to the schema which otherwise are not allowed." displayLabel="Config Attributes" xmlns="http://www.bentley.com/schemas/Bentley.ECXML.3.2">
```
ColinKerr, line 6: How about we call it 'SchemaUpgradeCustomAttributes'

Standard/Released/ConfigAttributes.01.00.00.ecschema.xml
```diff
@@ -0,0 +1,11 @@
...
+    <ECCustomAttributeClass typeName="AllowUnitChange" description="Applied to a property to indicate that existing KOQ can be changed with a KOQ having a different persistence unit." displayLabel="AllowUnitChange" appliesTo="PrimitiveProperty" modifier="Sealed">
```
ColinKerr, line 7: Do not set the display label to the same thing as the type name.  It's just a waste of bytes because we default to the typeName if the display label is not set.  In this case no need to set one at all because it's a custom attribute class so won't really show up in the UI.

Also for the description change: 'existing KOQ can be change ~~with~~ **to** a KOQ'

ColinKerr:
Also please write a remarks.md file for this schema with an examples of the CA being applied and the behavior.

jchick-bentley (approved):

jddecarlo (approved):
This is a nice addition. The Civil Connector has a couple of old errors in its schema where properties that should have been a ratio unit accidentally got assigned a length unit. The Civil Connector will be a nice test for this new support.

diegoalexdiaz (approved):

jonathandecarlo (approved):

ColinKerr (approved):

PR #93, opened 2022-05-25
Move unreleased schemas that are not expected to be worked to Unused folders

diegoalexdiaz:


karolis-zukauskas:
Diego, I think you can also move `GeoModeler` schema - it is an application specific schema for GeoModeler (ACDP) which was cancelled a while ago.

jchick-bentley (dismissed):

ColinKerr (approved):

PR #101, opened 2022-06-14
Upgrade schemas to ECXML 3.2 - batch #1.

diegoalexdiaz:


NabilSleiman (approved):
Diego communicated these changes. Thanks.

diegoalexdiaz:
> it seems like you added the wrong file? Domains/2-DisciplinePhysical/StructuralPhysical/StructuralPhysical.01.01.00.ecschema.xml.bak
> 
> above has the .BAK file extension

diegoalexdiaz:
@JValiunas , there is a longer story behind the inclusion of a bak file in this PR:

- The existing WIP StructuralPhysical schema in master (v1.1.0) is not ready to be released in the timeframe this upgrade to ECXML3.2 calls for. Thus, I had to rename it and copy the last released version (v1.0.0) as the new WIP (v1.0.1), which is the version upgraded in this PR.
- There is a bug in the schema-validation code when two ecschema.xml files with the same name are located in the same folder. It was leading to an error, preventing this PR from passing all its checks. The only workaround for now was to rename the v1.1.0 schema file (to .bak).

JValiunas (approved):

PR #106, opened 2022-06-23
Introduce PhysicalType classes to StructuralPhysical

diegoalexdiaz:


NabilSleiman (dismissed):

jchick-bentley (approved):

jchick-bentley (approved):

jchick-bentley (commented):
@diegoalexdiaz we have StructuralComponents which have StructuralComponentTypes which is how we intended to assign materials.  Is the intention here to assign the materials to a wall for example via the StructuralMemberType?

diegoalexdiaz:
@jchick-bentley, the expected pattern (described [here](https://www.itwinjs.org/bis/guide/physical-perspective/physical-materials/)) is a PhysicalMaterial instance referenced by either a PhysicalType (in which case, that material applies to all instances of such Type), or referenced by PhysicalElement instances directly (because they don't have a type-instance, or they are overriding the material specified by their type-instance).

In the case of a complex wall like the one from Revit, the Wall is modeled via an ArchPhys:WallType with layers. Those layers, also PhysicalTypes, are composed by the WallType in a particular order and thicknesses. Those "layerTypes" can reference their PhysicalMaterial.

The ArchPhys schema already has a WallLeafType class which applies to some of the wall-layers that are modeled in Revit. When a wall-layer in Revit is understood as Structural, according to what I've understood from you guys, the Revit Connector can then map it to a StructuralPhysical:WallType and reference the appropriate PhysicalMaterial for it.

In summary, the physicalMaterial is then specified per layer-type in the composite Wall case, as opposed to being assigned to the overall "wall-type" for the entire composite.

jchick-bentley:
> @jchick-bentley, the expected pattern (described [here](https://www.itwinjs.org/bis/guide/physical-perspective/physical-materials/)) is a PhysicalMaterial instance referenced by either a PhysicalType (in which case, that material applies to all instances of such Type), or referenced by PhysicalElement instances directly (because they don't have a type-instance, or they are overriding the material specified by their type-instance).
> 
> In the case of a complex wall like the one from Revit, the Wall is modeled via an ArchPhys:WallType with layers. Those layers, also PhysicalTypes, are composed by the WallType in a particular order and thicknesses. Those "layerTypes" can reference their PhysicalMaterial.
> 
> The ArchPhys schema already has a WallLeafType class which applies to some of the wall-layers that are modeled in Revit. When a wall-layer in Revit is understood as Structural, according to what I've understood from you guys, the Revit Connector can then map it to a StructuralPhysical:WallType and reference the appropriate PhysicalMaterial for it.
> 
> In summary, the physicalMaterial is then specified per layer-type in the composite Wall case, as opposed to being assigned to the overall "wall-type" for the entire composite.

This is the first time I looked at the schema since the topic arose.  After I approved it, I saw all the component types we added a few years ago and that refreshed my memory.  A StructuralMember is derived from a StructuralAssembly - Assembly of what? components.  The component types were added for the assignment of materials.  When we implemented components, that was the intent.  So a precast wall may have a concrete rectangular solid as its main component, but, may also have embedded steel angles for welding to structural steel which would be modeled as separate profiled components.  Material assignment by component via ComponentType rather than by WallType allows the for the clip angles to be part of the wall assembly and be assigned the correct material.

Another thing we need to consider for this specific revit issue, is that it concerns CMU block.  We could arguably assign that concrete, but, we probably should add separate support for CMU alongside Concrete, Steel, Aluminum etc.

jchick-bentley (dismissed):

Domains/2-DisciplinePhysical/StructuralPhysical/StructuralPhysical.ecschema.xml
```diff
@@ -34,31 +34,139 @@
     <ECEntityClass typeName="Slab" displayLabel="Slab">
         <BaseClass>StructuralMember</BaseClass>
     </ECEntityClass>
+    <ECEntityClass typeName="SlabType" displayLabel="Slab Type">
+        <BaseClass>bis:PhysicalType</BaseClass>
+    </ECEntityClass>
+    <ECRelationshipClass typeName="SlabIsOfType" strength="referencing" modifier="None" description="A type-instance relation; one that indicates that the specific sp:Slab is an instance of the defined sp:SlabType.">
```
laurynas111, line 52: Could we reuse `bis:PhysicalElementIsOfType` relationship for mapping newPhysicalType to PhysicalElements instead?
What advantages do you see of having separate relationships for each type?

diegoalexdiaz: These sub-classes of bis:PhysicalElementIsOfType are primarily for documentation purposes. That is, it is easier for consumers of the schema to realize what PhysicalType-subclass is meant to be associated with what PhysicalElement-subclass.

rgonzalo-orellana (approved):
Approved – no issues found.

jchick-bentley (commented):
@diegoalexdiaz Did we revert all of the structural component classes and types - I don't see them anywhere?  The mixin IStructuralComponent and associated relationships such as StructuralAssemblyOwnsStructuralComponents and StructuralAssemblyUsesMainComponent are still in StructuralPhysicalInterop.ecschema.xml . IStructuralComponent is also used in Fasteners.ecschema.xml

diegoalexdiaz:
@jchick-bentley - the focus of this PR is about adding PhysicalTypes to both the version in production and the "next-version" you worked on back then. This PR also brings the very first remarks file documenting the version in production. The "next-version" (still temporarily called v1.1.0) is in the same folder "on the side". This PR only renames it to "ecschema.xml" now that the schema-check tools in github can handle two WIP schemas in the same folder. Before it has been kept around as ".xml.bak". All its original content is still in there... it just has new PhysicalType classes.

jchick-bentley (approved):

BinithAnto (approved):

PR #107, opened 2022-06-23
Add the NotSubclassableInReferencingSchemas CA to key base Core classes.

diegoalexdiaz:


caseymullen (commented):
Those three additions all make logical sense to me. It would be interesting to test what happens when we try to import the tower schema after this change. How does it fail?  Causing some soft failure that will be flagged in development but would crash a process would perhaps be a good compromise

PR #108, opened 2022-06-27
GeologicalModel schema

diegoalexdiaz:


Domains/4-Application/Connectors/GeologicalModel/GeologicalModel.ecschema.xml
```diff
@@ -0,0 +1,46 @@
...
+    <ECEntityClass typeName="GeologicalModel" modifier="Sealed" description="An AnalyticalModel specialization expected to contain GeologicElement3d instances at different granularities.">
+        <BaseClass>anlyt:AnalyticalModel</BaseClass>
+    </ECEntityClass>
+
+    <ECEntityClass typeName="GeologicElement3d" modifier="Abstract" description="An AnalyticalElement serving as common base class for all element-classes modeling Geological concepts.">
+        <BaseClass>anlyt:AnalyticalElement</BaseClass>
+    </ECEntityClass>
+
+    <ECEntityClass typeName="FaultBlock" modifier="None" description="A GeologicElement3d modeling a set of volumes and surfaces that may be partitioned by one or more Fault Surfaces">
+        <BaseClass>GeologicAnalyticalPartition</BaseClass>
```
danywu, line 20: @diegoalexdiaz the `FaultBlock` itself does not have any geometry associated. It's simply a grouping of `GeologicElement3D` instances. Would it make sense to inherit from the `GeologicAnalyticalPartition` instead?

diegoalexdiaz: @danywu no, it wouldn't make sense. The purpose of InformationPartition classes in BIS is to advertise the modeling-perspective that a branch below a Subject focuses on. The GeologicAnalyticalPartition class is doing just that, so it is left 'sealed'. For more info on the purpose of InformationPartitions, see https://ddiazbisdocs.z13.web.core.windows.net/bis/guide/data-organization/modeling-perspectives/

On the other hand, a FaultBlock is modeling a specific concept in your domain. Thus, it needs to be a subclass of AnalyticalElement (via GeologicElement3d). It doesn't matter if such element typically doesn't have geometry on its own. It is fine that their `GeometryStream` property is left NULL.

I think the "Partition" wording of the InformationPartition classes in BIS may be introducing confusion in your domain as such word has a strong meaning in your world. Note that InformationPartition classes are typically not shown in the UX so they won't generate confusion to end-users. Their purpose is rather aimed at software-discoverable semantics (e.g. a piece of software can traverse the Subject hierarchy and find the InformationPartitions that it is designed to handle - e.g. Physical or Documents or Definitions or a specific kind of Analysis).

Domains/4-Application/Connectors/GeologicalModel/GeologicalModel.ecschema.xml
```diff
@@ -0,0 +1,46 @@
...
+    <ECEntityClass typeName="GeologicalModel" modifier="Sealed" description="An AnalyticalModel specialization expected to contain GeologicElement3d instances at different granularities.">
+        <BaseClass>anlyt:AnalyticalModel</BaseClass>
+    </ECEntityClass>
+
+    <ECEntityClass typeName="GeologicElement3d" modifier="Abstract" description="An AnalyticalElement serving as common base class for all element-classes modeling Geological concepts.">
+        <BaseClass>anlyt:AnalyticalElement</BaseClass>
+    </ECEntityClass>
+
+    <ECEntityClass typeName="FaultBlock" modifier="None" description="A GeologicElement3d modeling a set of volumes and surfaces that may be partitioned by one or more Fault Surfaces">
+        <BaseClass>GeologicElement3d</BaseClass>
```
danywu, line 20: What's the best way to associate `GeologicalCategoryVolume` and `ContactSurface` instances to `FaultBlock` instances? Volumes and Contact Surfaces "belong" to a particular Fault Block.

diegoalexdiaz: That's what the sentence in the remarks address:

"_Instances of `GeologicalCategoryVolume` are expected to be contained in a submodel of a `FaultBlock` instance that it is part of._"

What that means is that whoever creates a `FaultBlock` instance (e.g. a Connector) needs to create a submodel for it. Then, instances of `GeologicalCategoryVolume` or `ContactSurface` associated to such `FaultBlock` needs to be created in the appropriate submodel.

For further illustration, consider the example at: https://www.itwinjs.org/learning/backend/createelements/#geometricelement3d ... which inserts an element in the iModel.

The equivalent code for a `GeologicalCategoryVolume` instance would look like:

```
const props = {
    model: faultBlockSubmodelId, // submodel created earlier, when the correspoinding FaultBlock instance was inserted into the iModel
    classFullName: "geologicalModel:GeologicalCategoryVolume",
    ...
  };
  return iModelDb.elements.insertElement(props);
```
In summary, BIS currently doesn't have a schema-based technique to enforce the kind of model expected for a particular element-class. Currently, the only way to add constrains/validation in such situation is via a "domain-handler" - Typescript code written for a particular schema, reacting to certain operations (e.g. OnInsert) so it has the opportunity to check any needed domain-specific conditions and fail the operation if they are not met. However, writing a "domain-handler" for a schema is not mandatory, so it is generally the data-writer's responsibility (i.e. your connector) to follow the expected element-organization documented in a schema (if any).

danywu: Ah, that explains it. Thanks, @diegoalexdiaz . The team is looking at using PCF for this connector. @jchick-bentley , is this technique supported in PCF?

jchick-bentley: @danywu I'm not aware implementations that use the ModelModelsElement pattern proposed here, but, at the same time I don't see any obvious limitations.  When I last discussed PCF, I said if you can picture your data model in tabular format like an excel file.  It is difficult to picture how this pattern could be represented in a single spread sheet, but, it the xlsx loader could be extended to accommodate multiple sheets and it is not the only loader option.  It's easier to see this pattern represented in JSON format and there's a JSON loader.  I'm optimistic we can get it to work.  I recommend scheduling this task ahead of others to identify any impediments early.

jchick-bentley (approved):

.github/CODEOWNERS
```diff
@@ -73,6 +73,7 @@ snapshotInformation.json
 
 /Domains/1-Common/Unused/                                                           @iTwin/bis-core
 /Domains/2-DisciplinePhysical/Unused/                                               @iTwin/bis-core
+/Domains/3-Application/Unused/                                                      @iTwin/bis-core
```
ColinKerr, line 76: This path doesn't exist.  Do you mean `/Domains/3-DisciplineOther/`

Domains/3-DisciplineOther/GeologicalModel/GeologicalModel.ecschema.xml
```diff
@@ -0,0 +1,76 @@
...
+
+    <ECEntityClass typeName="GeologicalCategoryVolume" modifier="None" description="A GeologicElement3d modeling a Volume of a particular Category. Categories commonly represent Lithology.">
+        <BaseClass>GeologicElement3d</BaseClass>
+    </ECEntityClass>
+
+    <ECEntityClass typeName="ContactSurface" modifier="None" description="A GeologicElement3d modeling a surface between two or more Category Volumes.">
+        <BaseClass>GeologicElement3d</BaseClass>
+    </ECEntityClass>
+
+    <ECRelationshipClass typeName="FaultBlockOwnsGeologicalCategoryVolume" modifier="None" strength="referencing" description="Relates a FaultBlock to child GeologicCategoryVolumes.">
```
jchick-bentley, line 42: @danywu this relationship should have strength = "embedding" if it is to inherit from "ElementOwnsChildElements".  I don't think it will pass the schema validation with strength="referencing"

Domains/3-DisciplineOther/GeologicalModel/GeologicalModel.ecschema.xml
```diff
@@ -0,0 +1,76 @@
...
+        <BaseClass>bis:ElementOwnsChildElements</BaseClass>
+        <Source multiplicity="(0..1)" roleLabel="owns" polymorphic="true">
+            <Class class="FaultBlock"/>
+        </Source>
+        <Target multiplicity="(0..*)" roleLabel="is owned by" polymorphic="true">
+            <Class class="GeologicalCategoryVolume"/>
+        </Target>
+    </ECRelationshipClass>
+
+    <ECRelationshipClass typeName="FaultBlockOwnsContactSurface" modifier="None" strength="referencing" description="Relates a FaultBlock to child ContactSurfaces.">
```
jchick-bentley, line 52: @danywu same as above. strength should be embedding.

danywu:
> @danywu same as above. strength should be embedding. 

You're right, Peter discovered this too. I will fix these up.

danywu (dismissed):

jchick-bentley (approved):

karolis-zukauskas (approved):

Domains/3-DisciplineOther/GeologicalModel/GeologicalModel.ecschema.xml
```diff
@@ -0,0 +1,55 @@
...
+    <ECEntityClass typeName="FaultBlock" modifier="None" description="A GeologicElement3d modeling a set of volumes and surfaces that may be partitioned by one or more Fault Surfaces">
+        <BaseClass>GeologicElement3d</BaseClass>
+        <BaseClass>bis:IParentElement</BaseClass>
+    </ECEntityClass>
+
+    <ECEntityClass typeName="GeologicalCategoryVolume" modifier="None" description="A GeologicElement3d modeling a Volume of a particular Category. Categories commonly represent Lithology.">
+        <BaseClass>GeologicElement3d</BaseClass>
+    </ECEntityClass>
+
+    <ECEntityClass typeName="ContactSurface" modifier="None" description="A GeologicElement3d modeling a surface between two or more Category Volumes.">
```
karolis-zukauskas, line 28: Where will the surface be stored? In `GeotechnicalInterpretation` we had binary properties on similar classes to store such data (e.g. meshes) in an optimized way. I assume that geom stream is going to be used now - are we sure this is the correct approach?

danywu: Yes, the geometry stream is where we are storing it currently, as IndexedPolyface. Our surfaces (in the source of the sync) are all triangle meshes, so the indexes are Nx3. I would love to hear more about this optimised way, specifically how and for which clients/purposes. At this stage we're going ahead with storing it in the GeometryStream, but we should have offline discussion on better ways to store this in iModel.

ColinKerr (approved):

PR #133, opened 2022-08-25
WiMSE icon added

naveedkhan8067:
The icon is required in media so that it can show up in the documentation.


ColinKerr:
I don't like that we are putting the icon in the core media folder when we are going to use it for all schemas.  It is odd that we're putting this icon in the bis schemas repo but I understand we need a place to put it.

I think the script will grab any media folder so long as it's not in the docs directory.  So maybe the right thing to do is to add a new media folder in the root.  What do you think @naveedkhan8067 and @diegoalexdiaz

diegoalexdiaz:
> I don't like that we are putting the icon in the core media folder when we are going to use it for all schemas. It is odd that we're putting this icon in the bis schemas repo but I understand we need a place to put it.
> 
> I think the script will grab any media folder so long as it's not in the docs directory. So maybe the right thing to do is to add a new media folder in the root. What do you think @naveedkhan8067 and @diegoalexdiaz

I share the odd feeling about this icon needed to go in the bis-schemas repo, even in a new media folder in the root. I don't know any implementation details behind the script but my initial expectation is for this icon to be stored in the sample place as the script. 

If such script is in the bis-schemas repo, could it go in the sub-folder where it is and be deployed to the right place by the script when it runs?

naveedkhan8067:
> I don't like that we are putting the icon in the core media folder when we are going to use it for all schemas. It is odd that we're putting this icon in the bis schemas repo but I understand we need a place to put it.
> 
> I think the script will grab any media folder so long as it's not in the docs directory. So maybe the right thing to do is to add a new media folder in the root. What do you think @naveedkhan8067 and @diegoalexdiaz

I was also first thinking to create a new media folder for that but I think at least its better than this.

naveedkhan8067:
> > I don't like that we are putting the icon in the core media folder when we are going to use it for all schemas. It is odd that we're putting this icon in the bis schemas repo but I understand we need a place to put it.
> > I think the script will grab any media folder so long as it's not in the docs directory. So maybe the right thing to do is to add a new media folder in the root. What do you think @naveedkhan8067 and @diegoalexdiaz
> 
> I share the odd feeling about this icon needed to go in the bis-schemas repo, even in a new media folder in the root. I don't know any implementation details behind the script but my initial expectation is for this icon to be stored in the sample place as the script.
> 
> If such script is in the bis-schemas repo, could it go in the sub-folder where it is and be deployed to the right place by the script when it runs?

That python script is actually part of our YAML file at: https://github.com/iTwin/bis-schemas/tree/master/tools/MarkdownGeneration/generate-docs.yaml but in this case the icon will become part of the tools folder and also I have to check whether this script  ignores this directory or not.

ColinKerr:
> > > I don't like that we are putting the icon in the core media folder when we are going to use it for all schemas. It is odd that we're putting this icon in the bis schemas repo but I understand we need a place to put it.
> > > I think the script will grab any media folder so long as it's not in the docs directory. So maybe the right thing to do is to add a new media folder in the root. What do you think @naveedkhan8067 and @diegoalexdiaz
> > 
> > 
> > I share the odd feeling about this icon needed to go in the bis-schemas repo, even in a new media folder in the root. I don't know any implementation details behind the script but my initial expectation is for this icon to be stored in the sample place as the script.
> > If such script is in the bis-schemas repo, could it go in the sub-folder where it is and be deployed to the right place by the script when it runs?
> 
> That python script is actually part of our YAML file at: https://github.com/iTwin/bis-schemas/tree/master/tools/MarkdownGeneration/generate-docs.yaml but in this case the icon will become part of the tools folder and also I have to check whether this script ignores this directory or not. What do you suggest to put it in that folder or create a media folder at root?

What if the image was included in the ecjson2md package and it copied the image to a media folder the same directory as the output md files

naveedkhan8067:
> > > > I don't like that we are putting the icon in the core media folder when we are going to use it for all schemas. It is odd that we're putting this icon in the bis schemas repo but I understand we need a place to put it.
> > > > I think the script will grab any media folder so long as it's not in the docs directory. So maybe the right thing to do is to add a new media folder in the root. What do you think @naveedkhan8067 and @diegoalexdiaz
> > > 
> > > 
> > > I share the odd feeling about this icon needed to go in the bis-schemas repo, even in a new media folder in the root. I don't know any implementation details behind the script but my initial expectation is for this icon to be stored in the sample place as the script.
> > > If such script is in the bis-schemas repo, could it go in the sub-folder where it is and be deployed to the right place by the script when it runs?
> > 
> > 
> > That python script is actually part of our YAML file at: https://github.com/iTwin/bis-schemas/tree/master/tools/MarkdownGeneration/generate-docs.yaml but in this case the icon will become part of the tools folder and also I have to check whether this script ignores this directory or not. What do you suggest to put it in that folder or create a media folder at root?
> 
> What if the image was included in the ecjson2md package and it copied the image to a media folder the same directory as the output md files

I think it should be possible. I will try this out.

naveedkhan8067:
> > > > > I don't like that we are putting the icon in the core media folder when we are going to use it for all schemas. It is odd that we're putting this icon in the bis schemas repo but I understand we need a place to put it.
> > > > > I think the script will grab any media folder so long as it's not in the docs directory. So maybe the right thing to do is to add a new media folder in the root. What do you think @naveedkhan8067 and @diegoalexdiaz
> > > > 
> > > > 
> > > > I share the odd feeling about this icon needed to go in the bis-schemas repo, even in a new media folder in the root. I don't know any implementation details behind the script but my initial expectation is for this icon to be stored in the sample place as the script.
> > > > If such script is in the bis-schemas repo, could it go in the sub-folder where it is and be deployed to the right place by the script when it runs?
> > > 
> > > 
> > > That python script is actually part of our YAML file at: https://github.com/iTwin/bis-schemas/tree/master/tools/MarkdownGeneration/generate-docs.yaml but in this case the icon will become part of the tools folder and also I have to check whether this script ignores this directory or not. What do you suggest to put it in that folder or create a media folder at root?
> > 
> > 
> > What if the image was included in the ecjson2md package and it copied the image to a media folder the same directory as the output md files
> 
> I think it should be possible. I will try this out.

I have added that logo in the same PR which was already open for logo related task in the ecjson2md repository: https://github.com/iTwin/ecjson2md/pull/10 

Also, I have updated the YAML pipeline here for copying the logo to the media directory.

ColinKerr (approved):

PR #149, opened 2022-10-21
Replaced certain beam segment properties that had the wrong unit with…

jonathandecarlo:
… their correct units. Also, I needed to define some new KoQ's in CifUnits for this purpose.


diegoalexdiaz:
I assume you're taking the deprecate/hide approach for these properties because the Custom-attribute way to change a property's KOQ to a different dimension didn't work?

jonathandecarlo:
Nope, I just forgot about that support. Thanks for the reminder. I'll push changes shortly to go that route.

jonathandecarlo:
@diegoalexdiaz I switched to using the schema update koq switch tags. The automatic checks seem to be passing now. I think it is ready for review again when you get the chance. Thanks!

diegoalexdiaz (approved):

PR #169, opened 2022-12-05
ISM schema Tendon API update

arnoldas1:
Updated ISM schema with latest ISM Tendon API changes.
Additionally, fixed boolean parameters to match xsd, removed empty fields and updated used referenced schemas.


naveedkhan8067 (commented):
/azp run

PR #236, opened 2023-09-12
1259630 Update CifGeometricRules.ecschema

Bordoloi123:
Updated CifGeometricRules.ecschema for removal of duplication of FilletRuleBaseAspect properties done by sub classes

### DIFF
--- a/Domains/4-Application/Connectors/CivilInfrastructureFramework/CifGeometricRules.ecschema.xml
+++ b/Domains/4-Application/Connectors/CivilInfrastructureFramework/CifGeometricRules.ecschema.xml
@@ -1401,7 +1401,7 @@
             <Class class="EqualSpacePointRuleAspect"/>
         </Target>
     </ECRelationshipClass>
-    <ECEntityClass typeName="FilletRuleBaseAspect" modifier="Abstract">
+	<ECEntityClass typeName="FilletRuleCommonAspect">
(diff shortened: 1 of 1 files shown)


Domains/4-Application/Connectors/CivilInfrastructureFramework/CifGeometricRules.ecschema.xml
```diff
@@ -1401,7 +1401,7 @@
             <Class class="EqualSpacePointRuleAspect"/>
         </Target>
     </ECRelationshipClass>
-    <ECEntityClass typeName="FilletRuleBaseAspect" modifier="Abstract">
+	<ECEntityClass typeName="FilletRuleCommonAspect">
```
diegoalexdiaz, line 1404: You can't delete (or rename) classes in a minor schema-change. You'll need to leave the class name the way it was while deprecating & hiding its properties. In parallel, you duplicate these properties into a new class with a different name.

Bordoloi123: The original base class FilletRuleBaseAspect is not deleted. It is there from line no. 1429. with all the properties hidden and deprecated.

diegoalexdiaz (approved):

PR #248, opened 2023-10-19
Release updated CifBridge, CifCommon & CifRail ecschema

Bordoloi123:


SchemaInventory.json
```diff
@@ -3090,14 +3090,28 @@
...
       "dynamic": "No",
       "approved": "Yes"
+    },
+    {
+      "name": "CifBridge",
+      "path": "Domains\\4-Application\\Connectors\\CivilInfrastructureFramework\\Released\\CifBridge.01.00.11.ecschema.xml",
+      "released": true,
+      "version": "01.00.11",
+      "comment": "",
+      "verifier": "",
```
ColinKerr, line 3100: you have to put a verifier, you can use my name as a verifier.

In the past we have not used the comment field but I would like to start using it so we have a better idea of why a schema was released.  So please put a sentence or two describing the change to each schema.  In this case it will be very simple

PR #280, opened 2024-02-28
Release Formats ecschema

Bordoloi123:
Added AngleDM format for use in Dynamic CifUnits ecschema


ColinKerr (approved):
I fixed spacing, there were tabs instead of spaces which can cause inconsistent formatting

PR #290, opened 2024-03-20
Release updated CifRail schema with new Rail properties

Bordoloi123:
The update includes new properties of  Turnout, Sleeper, Equipment , etc.

### DIFF
--- a/Domains/4-Application/Connectors/CivilInfrastructureFramework/CifRail.ecschema.xml
+++ b/Domains/4-Application/Connectors/CivilInfrastructureFramework/CifRail.ecschema.xml
@@ -1,4 +1,8 @@
 <?xml version="1.0" encoding="UTF-8"?>
+<!-- ==================================================================================
+|  * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+|  * See LICENSE.md in the project root for license terms and full copyright notice.
+======================================================================================= -->
 <ECSchema schemaName="CifRail" alias="cifrl" version="01.00.08" description="iModel Connector schema containing aspect classes with properties from OpenRail Designer." xmlns="http://www.bentley.com/schemas/Bentley.ECXML.3.2">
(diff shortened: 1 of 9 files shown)


Domains/4-Application/Connectors/CivilInfrastructureFramework/CifRail.ecschema.xml
```diff
@@ -3,20 +3,37 @@
...
+        <ECEnumerator name="Outer" value="2" displayLabel="Outer"/>
+    </ECEnumeration>
+    <ECEnumeration typeName="BaseTurnoutObjectSettingsAspect_BaseTurnoutObjectSettings_NumberOfBranches_Enum" backingTypeName="int" isStrict="true">
+        <ECEnumerator name="One" value="1" displayLabel="1"/>
+        <ECEnumerator name="Two" value="2" displayLabel="2"/>
+    </ECEnumeration>
+    <ECEnumeration typeName="CableAspect_Cable_ShowCable_BoolEnum" backingTypeName="int" isStrict="true">
+        <ECEnumerator name="True" value="1" displayLabel="Yes"/>
+        <ECEnumerator name="False" value="0" displayLabel="No"/>
+    </ECEnumeration>
```
ColinKerr, line 36: This isn't a great pattern, I see this isn't a new pattern (there are many existing enums with this pattern) so I won't insist that it be changed but it should be changed.  Once released it's a breaking change to change the type of the enum even if it is functionally the same, this is because it could break existing ECSql that references the enum

Bordoloi123: This was generated by a tool we use (cif2bis) and is output of a properties service( in civilplatform) written to get all the presentation classes from Civil products. In case we change the pattern manually, next time we run the tool, this will be again updated to the existing pattern. If this needs to handled , can you suggest the pattern that is favorable?

ColinKerr (approved):

PR #314, opened 2024-06-13
Switch KOQs over to RoadRailUnits schema.

diegoalexdiaz:


BenTWhite (approved):
easy fix - looks good

PR #316, opened 2024-06-14
Rename "Standard" folder into "System".

diegoalexdiaz:


ColinKerr:
You also need to update:
 - CODEOWNERS
 - tools/MarkdownGeneration/generate-docs.yaml
 - tools/SchemaValidation/validateSchemas.yaml (currently commented out but we should keep things clean

ColinKerr (approved):

PR #328, opened 2024-07-12
Adding the "Required" property to connector:RuleViolation

jffmarker:
In the final step of updating a model, the connector now detects and deletes rule violations involving elements that a) were skipped by a prior update and b) that were caused by external elements that have since been deleted. To support this, I added the "Required" property to connector:RuleViolation. See the description in the schema.

### DIFF
--- a/Domains/4-Application/Connectors/Connector.ecschema.xml
+++ b/Domains/4-Application/Connectors/Connector.ecschema.xml
@@ -3,13 +3,14 @@
 |  * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
 |  * See LICENSE.md in the project root for license terms and full copyright notice.
 ======================================================================================= -->
-<ECSchema schemaName="Connector" alias="connector" version="01.00.00" description="ECClasses used by all connectors" xmlns="http://www.bentley.com/schemas/Bentley.ECXML.3.2">
+<ECSchema schemaName="Connector" alias="connector" version="01.01.00" description="ECClasses used by all connectors" xmlns="http://www.bentley.com/schemas/Bentley.ECXML.3.2">
     <ECSchemaReference name="BisCore" version="01.00.13" alias="bis" />
     <ECSchemaReference name="CoreCustomAttributes" version="01.00.03" alias="CoreCA"/>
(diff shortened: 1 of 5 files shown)


ColinKerr (approved):

Domains/4-Application/Connectors/Connector.ecschema.xml
```diff
@@ -3,13 +3,14 @@
 |  * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
 |  * See LICENSE.md in the project root for license terms and full copyright notice.
 ======================================================================================= -->
-<ECSchema schemaName="Connector" alias="connector" version="01.00.00" description="ECClasses used by all connectors" xmlns="http://www.bentley.com/schemas/Bentley.ECXML.3.2">
+<ECSchema schemaName="Connector" alias="connector" version="01.01.00" description="ECClasses used by all connectors" xmlns="http://www.bentley.com/schemas/Bentley.ECXML.3.2">
```
diegoalexdiaz, line 6: Since this PR only changes the WIP copy of the schema, there's no need to modify its version number.

jffmarker: There are already iModels that use this schema, so we need some way to know we need an upgrade I believe...

diegoalexdiaz: Typically schema-version increases happen when a schema is released. WIP schemas shouldn't be imported into production iModels. If a WIP schema ends up being imported into a production iModel (typically by mistake), we should capture such schemas as "released", then increment its version from it.

1. Note that schemas now advertise their "status" (WIP schemas always say they are "not for production") - I thought there were checks in Connector pipelines against releasing schemas with a status other than "for production".
2. If you really need to increase the version of a schema, please be aware of the meaning of the numbers in the version since they communicate what kind of changes are introduced: [link](https://www.itwinjs.org/bis/guide/schema-evolution/schema-versioning-and-generations/#schema-version-numbers). In this PR's case, the next version number that would capture the kind of change introduced would be 1.0.1.

diegoalexdiaz (approved):

diegoalexdiaz (approved):

ColinKerr (approved):

PR #339, opened 2024-08-19
Horizontal Direction Units

diegoalexdiaz:


rschili (approved):

System/Units.ecschema.xml
```diff
@@ -87,6 +87,7 @@
     <Phenomenon typeName="MOLE" definition="MOLE" displayLabel="Mole" />
     <Phenomenon typeName="LUMINOSITY" definition="LUMINOSITY" displayLabel="Luminosity" />
     <Phenomenon typeName="ANGLE" definition="ANGLE" displayLabel="Angle" />
+    <Phenomenon typeName="HORIZONTAL_DIRECTION" definition="HORIZONTAL_DIRECTION" displayLabel="Horizontal Direction" />
```
rschili, line 90: I may be confused, but now that we are using a North-Base, isn't it a "Vertical Direction"?

rschili: Ah no I see, horizontal in the sense that it's planar without elevation. That makes sense. Closing the comment.

rschili (approved):

ColinKerr (approved):

PR #343, opened 2024-08-27
Bearing Formats

diegoalexdiaz:


System/Formats.ecschema.xml
```diff
@@ -50,4 +50,12 @@
             <Unit label="sec">u:S</Unit>
         </Composite>
     </Format>
+
+    <Format typeName="BearingDMS" displayLabel="Degrees Minutes Seconds" type="Bearing" precision="0" uomSeparator="" revolutionUnit="u:BEARING_REVOLUTION" formatTraits="showUnitLabel">
```
rschili, line 54: Oh I just noticed this PR. This is not going to work at the moment. The new formats support is just built into the core-quantity package and native units framework. Supporting it inside an ECSchema is a different deal. @ColinKerr: I think this may even require Rohit's backwards-compatibility changes to roll out before we can add something like this, as the new type would not be handled by older code.

PR #361, opened 2024-09-19
adding schema for design pipe sizes

BenTWhite:
Adding two new Definition elements to allow users to define which shapes and materials to use in the automated design.

- Pipe Size Set - A set of pipe sizes with the same shape and material.... and owns a set of Pipe Sizes
- Pipe Size - gives a selectable unique id for each pipe size along with an initial roughness value...
-  Each pipe size points to a PipeShapeAspect which already exists


diegoalexdiaz:
Ben and I discussed this topic offline. The approach he is introducing here is likely to be temporary because it creates simplified duplicates of concepts already captured by other BIS patterns (i.e. physical material and pipingPort Shape & Size).

I'm approving this PR in order to get his team moving forward with additional implementation details, but we'll discuss this topic further in the near future.

diegoalexdiaz (approved):

PR #377, opened 2024-10-29
Outlet Structures and Outfalls

BenTWhite:
1. Added DrainageNode analytical element - to make relationships betweeen drainage elements less generalized
2. Added Conduit analytical element  for the same reason - eventually will have a channel for subclass.

**Outlet Structures - Structures which control the flow from a node like a pond into an edge like a pipe or channel**
1. Added FlowControlStructure - which has the attribute for flow direction.
2. The FlowControlStructure references a node (generally a pond, but ultimately be other structures)
4. The FlowControlStructure references a link (for example a pond with a spillway weir that empties into a channel)
5. The concrete FlowControlStructures are PondRiser, Weir, and Orifice

**Outfalls**
1. Added BoundryConditionAspect to Outfalls
2  The BoundaryConnectionAspect which subclasses the BoundryConditionAspect references a DrainageNode


Domains/3-DisciplineOther/Hydraulics/SewerHydraulicAnalysis.ecschema.xml
```diff
@@ -168,6 +204,142 @@
...
+      <Class class="Orifice"/>
+    </Source>
+    <Target multiplicity="(1..1)" roleLabel="owned by" polymorphic="true">
+      <Class class="OrificeShapeAspect"/>
+    </Target>
+  </ECRelationshipClass>
+  
+	<ECEntityClass typeName="CircularOrificeAspect" modifier="Sealed" displayLabel="Circular Orifice">
+    <BaseClass>OrificeShapeAspect</BaseClass>
+    <ECProperty propertyName="Diameter" typeName="double" displayLabel="Diameter" category="HydraulicData" kindOfQuantity="rru:LENGTH" description="The diameter of the orifice opening"/>
```
diegoalexdiaz, line 288: It seems to me that the "LENGTH" KoQ may not be the correct one for "Diameter" properties. Such KoQ is meant to use larger units by default (e.g. Meters or Feet). There is another KoQ in RoadRailUnits called "LENGTH_SHORT" which seems more appropriate.

Having said that, I can tell nobody has used the latter KoQ since its default presentation units are currently wrong (Meters and Feet-Inches). They should probably be mm and inches.

diegoalexdiaz: @BenTWhite will address this detail in the next PR.

diegoalexdiaz (approved):

PR #378, opened 2024-10-30
Adding SWMM calculation options - along with some minor unit tweaks

BenTWhite:
1. I made a pass on diameters and changed the koq to LENGTH_SHORT where appropriate - structure diameter is usually presented in the larger units for example
2.  There were a couple of spots using the wrongs units , or using the AECU units instead of the rru
3. Add a enum to the rainfalleventset to store the type of rainfall data associated with it
4. refactor the CalcOptions so they are more like Design constraints...

There is now a CalculationOptions element with an enum for engine type, and based on the engine type will select the appropriate calc options aspect


diegoalexdiaz:
Thanks for addressing KoQs on various properties of this schema. Since the presentationUnits on the LENGTH_SHORT KoQ (RoadRailUnits schema) are currently wrong (i.e. Meters for SI, Feet-Index for US), do you want to address that in this PR? Or in a separate one?

diegoalexdiaz (approved):

PR #381, opened 2024-11-06
Support the labeling of the Cant Station Points

SachBhosale:
Cant station points are shown with indices as labels, but now with this correction, they will be displayed with their station.


diegoalexdiaz:
Bentley_Standard_CustomAttributes is only supported in PowerPlatform, not in iTwin.js. What you're aiming to achieve with this PR (i.e. driving the UX to use a particular property as the "Label" of an Element) can only be done via PresentationRules in iTwin.js. That is, the Connector needs to write a "Ruleset" element containing the needed PresentationRule, into its private DefinitionModel. No need to change any schemas to achieve this behavior.

diegoalexdiaz:
BTW, this is the schema and class to use when storing a "Ruleset" in an iModel: https://github.com/iTwin/bis-schemas/blob/master/Domains/4-Application/PresentationRules/Released/PresentationRules.01.00.01.ecschema.xml

SachBhosale:
We have managed this issue through code and no schema changes are necessary so I have closed this pull request.

PR #392, opened 2024-11-27
Initial commit for the new CivilSpatial schema.

diegoalexdiaz:


Domains/2-DisciplinePhysical/Civil/SpatialComposition/CivilSpatial/CivilSpatial.ecschema.xml
```diff
@@ -0,0 +1,135 @@
...
+    <ECRelationshipClass typeName="ParkingRowIsOfType" strength="referencing" modifier="None" description="A type-instance relation; one that indicates that the specific cvsp:ParkingRow is an instance of the defined cvsp:ParkingRowType.">
+        <BaseClass>spcomp:SpaceIsOfType</BaseClass>
+        <Source multiplicity="(0..*)" roleLabel="is of" polymorphic="true">
+            <Class class="ParkingRow" />
+        </Source>
+        <Target multiplicity="(0..1)" roleLabel="is type of" polymorphic="true">
+            <Class class="ParkingRowType"/>
+        </Target>
+    </ECRelationshipClass>
+    <ECEntityClass typeName="Sidewalk" displayLabel="Sidewalk" description="A cvsp:Space modeling a footpath for pedestrians.">
```
karolis-zukauskas, line 89: Some class definitions do not specify `modifier` attribute (like Sidewalk) while most other do. Perhaps we should specify it for every class for consistency?

Domains/2-DisciplinePhysical/Civil/SpatialComposition/CivilSpatial/CivilSpatial.ecschema.xml
```diff
@@ -0,0 +1,135 @@
...
+            <SupportedUse>NotForProduction</SupportedUse>
+        </ProductionStatus>
+        <SchemaLayerInfo xmlns="BisCustomAttributes.01.00.00">
+            <Value>DisciplinePhysical</Value>
+        </SchemaLayerInfo>
+    </ECCustomAttributes>
+
+    <ECEntityClass typeName="Site" modifier="None" description="The spatial location of a geographic site on which Facilities are situated. Its geometry is typically provided as a 2D/Plan-projected boundary.">
+        <BaseClass>spcomp:Site</BaseClass>
+    </ECEntityClass>
```
karolis-zukauskas, line 25: nitpick:
Personally, having a single new line after class definitions makes it easier to read schema xml - it feels less cramped and you can more clearly see definitions boundaries.

Domains/2-DisciplinePhysical/Civil/SpatialComposition/CivilSpatial/CivilSpatial.ecschema.xml
```diff
@@ -0,0 +1,156 @@
...
+    <ECCustomAttributes>
+        <ProductionStatus xmlns="CoreCustomAttributes.01.00.00">
+            <SupportedUse>NotForProduction</SupportedUse>
+        </ProductionStatus>
+        <SchemaLayerInfo xmlns="BisCustomAttributes.01.00.00">
+            <Value>DisciplinePhysical</Value>
+        </SchemaLayerInfo>
+    </ECCustomAttributes>
+
+    <ECEntityClass typeName="Site" modifier="None" description="The spatial location of a geographic site on which Facilities are situated. Its geometry is typically provided as a 2D/Plan-projected boundary.">
```
ColinKerr, line 23: Why do we need specializations of Site and SiteType?  I get having a subclass of Space because we are adding properties.

it feels a bit uncomfortable to have derived classes with the same name as their base class.  I don't have a good suggestion for better names, CivilSite seems worse, but this does seem wrong.

diegoalexdiaz: Those specializations are needed because the Site and SiteType classes were introduced as abstract at the underlying SpatialComposition schema. I wasn't involved in that decision back then but I take it as the expectation that was introduced was to always require a specialization at each domain that implements it. 

The trouble comes from the fact that the concept of a "Site" in its more pure way is implemented by "the Civil vertical" in the way Bentley classifies Infrastructure project. Thus, I also faced the same discontent with the resulting identical class-names - I believe the choosing of the class-names for the base/abstract classes is more the problem: It could've been something like `BaseSite` or `SiteElement` (i.e. following a "naming rule" that we used to follow about appending "Element" to abstract classes), but, in my opinion, the term `Site` makes more sense semantically in the CivilSpatial schema than in the underlying SpatialComposition.

BTW, I also had the same discomfort about introducing another `Space` class in CivilSpatial - it wasn't about introducing new properties, but just overriding the KoQ of base properties.

ColinKerr: I can agree that naming of the base classes like SiteElement and SiteTypeElement would have been the right thing to do, but that time has passed.  Barring someone coming up with a quality name for these derived classes I think leaving them this way is fine.

Adding property overrides is a reasonable reason to create a derived class.  If we required that a property override listed its base property it would be harder to move properties down the class hierarchy w/o breaking existing schemas ... unless we had some complex rule that took the version of the reference into account when enforcing the tag.  I think a better option is to get the schema explorer to a better place where we can more easily use it to view an arbitrary set of schemas.

ColinKerr (approved):

karolis-zukauskas (approved):

karolis-zukauskas (approved):

PR #410, opened 2024-12-11
Add RepositoryInfoAspect and ChannelRootAspect.Version to capture "Source-Editor Info" for both Connectors and Authoring.

diegoalexdiaz:


Domains/0-Core/BisCore.remarks.md
```diff
@@ -396,7 +396,12 @@ Applications create channels to define portions of the model-hierarchy that they
...
+Applications can optionally use the `ChannelRootAspect.Version` property in order to assign a _semantic version_ number to the data-organization in a *channel*. _Semantic versioning_ aims at efficiently advertising read and write backwards-compatibility with respect to previous versions of the owning Application of a *channel*. 
+
+_Semantic versioning_ of *channels* follows the "[read-version].[write-version].[minor-version]" format.
+- [read-version] is incremented when the data-organization of a *channel* is changed in ways that older versions of an Application cannot safely understand it for reading purposes.
+- [write-version] is incremented when the data-organization of a *channel* is changed in ways that it is not safe for older versions of an Application to add, modify or delete data in it.
+- [minor-version] is incremented when the data-organization of a *channel* is changed in ways that older versions of an Application can still safely read it or write to it.
```
ColinKerr, line 404: Should we document how an application should update the version?

For example what should newer software do when it encounters a channel with an older patch version?  how about older software encounters a newer patch version.  Should the newer software always update and older software never update the channel version

diegoalexdiaz: I agree we should, but the question is whether the schema remarks are the right place vs. the [separate article](https://www.itwinjs.org/learning/backend/channel/) about Channels in iTwin.js. I'm inclined to explain those details in the latter article and add a link to it from these remarks. Do you have a different view?

diegoalexdiaz: @ColinKerr, I addressed your question in the new examples of semantic versioning in the [Channel article](https://github.com/iTwin/itwinjs-core/pull/7558) referenced by the remarks.

ColinKerr (approved):

PR #414, opened 2024-12-19
Channels and end walls

BenTWhite:
1. fixed the properties on the curb inlet
2. add a channel element along with channel shape aspects
3. Added headwalls -  

this includes an enumeration with wall the InletControl coefficients

along with various aspects that describe the different ways headwalls can be treated: Outfall, PondInflow, PondOutflow, Transition

### DIFF
--- a/Domains/3-DisciplineOther/Hydraulics/SewerHydraulicAnalysis.ecschema.xml
+++ b/Domains/3-DisciplineOther/Hydraulics/SewerHydraulicAnalysis.ecschema.xml
@@ -211,6 +211,59 @@
         <ECEnumerator value="3" name="Slot" displayLabel="Slot Inlet"/>
     </ECEnumeration>
 
+    <ECEnumeration typeName="CurbThroatType" backingTypeName="int" isStrict="true" >
+        <ECEnumerator value="0" name="Horizontal" displayLabel="Horizontal"/>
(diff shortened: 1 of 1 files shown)


diegoalexdiaz:
Since version 1.0.0 was already released and the goal is to preserve compatibility with it going forward, that means that only minor schema changes are allowed. That is, changes that don't break any existing ECSQL queries written against the released version (i.e. read-compatibility).

Any deletion of a class or property is considered a major schema-update.  This PR is deleting some schema content from the released schema: e.g.

- CatchBasinTypeCurbInletHydraulicAspect.SlotWidth
- CatchBasinTypeCurbInletHydraulicAspect.SlotLength
- RectangularAspect class

In cases like these, you need to keep those properties/classes, but you can mark them "deprecated" and also hide them from the UX.

e.g.

```
<ECProperty propertyName=...>
  <ECCustomAttributes>
    <HiddenProperty xmlns="CoreCustomAttributes.01.00.03"/>
    <Deprecated xmlns="CoreCustomAttributes.01.00.03">
      <Description>TODO: Point existing readers of the deprecated property or class to how they can find the same data going forward. E.g. The new property or class.</Description>
    </Deprecated>
  </ECCustomAttributes>
</ECProperty>
```

diegoalexdiaz (approved):

PR #417, opened 2025-01-11
New enums and property type modification for Cant Station Point

SachBhosale:
This is to correct the properties of Cant Station Point.

### DIFF
--- a/Domains/4-Application/Connectors/CivilInfrastructureFramework/CifRail.ecschema.xml
+++ b/Domains/4-Application/Connectors/CivilInfrastructureFramework/CifRail.ecschema.xml
@@ -76,26 +76,6 @@
         <ECEnumerator name="ProjectedCant" value="17" displayLabel="Projected Cant"/>
         <ECEnumerator name="ProjectedCantAndLocation" value="18" displayLabel="Projected Cant And Location"/>
     </ECEnumeration>
-    <ECEnumeration typeName="CantStationPointAspect_CantStationPoint_CardinalPoint_Enum" backingTypeName="int" isStrict="true">
-        <ECEnumerator name="Undefined" value="-1" displayLabel="Undefined"/>
-        <ECEnumerator name="Start" value="0" displayLabel="Start"/>
-        <ECEnumerator name="End" value="1" displayLabel="End"/>
-        <ECEnumerator name="TangentToSpiral" value="2" displayLabel="Tangent To Spiral"/>
-        <ECEnumerator name="SpiralToCurve" value="3" displayLabel="Spiral To Curve"/>
-        <ECEnumerator name="CurveToSpiral" value="4" displayLabel="Curve To Spiral"/>
-        <ECEnumerator name="SpiralToTangent" value="5" displayLabel="Spiral To Tangent"/>
-        <ECEnumerator name="SpiralToSpiral" value="6" displayLabel="Spiral To Spiral"/>
-        <ECEnumerator name="TangentToCurve" value="7" displayLabel="Tangent To Curve"/>
-        <ECEnumerator name="CurveToTangent" value="8" displayLabel="Curve To Tangent"/>
-        <ECEnumerator name="CurveToCurve" value="9" displayLabel="Curve To Curve"/>
-        <ECEnumerator name="TangentToTangent" value="10" displayLabel="Tangent To Tangent"/>
-        <ECEnumerator name="GapToTangent" value="11" displayLabel="Gap To Tangent"/>
-        <ECEnumerator name="GapToSpiral" value="12" displayLabel="Gap To Spiral"/>
-        <ECEnumerator name="GapToCurve" value="13" displayLabel="Gap To Curve"/>
-        <ECEnumerator name="PointOnTangent" value="14" displayLabel="Point On Tangent"/>
-        <ECEnumerator name="PointOnCurve" value="15" displayLabel="Point On Curve"/>
-        <ECEnumerator name="PointOnSpiral" value="16" displayLabel="Point On Spiral"/>
-    </ECEnumeration>
     <ECEnumeration typeName="CantEntityAspect_CantEntity_RotateAbout_Enum" backingTypeName="int" isStrict="true">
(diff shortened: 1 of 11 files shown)


Domains/4-Application/Connectors/CivilInfrastructureFramework/CifRail.ecschema.xml
```diff
@@ -326,22 +388,33 @@
     </ECEntityClass>
     <ECEntityClass typeName="CantStationPointAspect" displayLabel="Cant Station Point">
         <BaseClass>cifcmn:CivilPresentation</BaseClass>
-        <ECProperty propertyName="CantStationPoint_TransitionType" typeName="string" displayLabel="Object Type" category="CantStationPoint_Cant_Station_Point" priority="299999"/>
-        <ECProperty propertyName="CantStationPoint_Station" typeName="double" displayLabel="Station" category="CantStationPoint_Cant_Station_Point" priority="299998" kindOfQuantity="cifu:STATION"/>
+        <ECProperty propertyName="CantStationPoint_TransitionType" typeName="CantStationPointAspect_CantStationPoint_TransitionType_Enum" displayLabel="Object Type" category="CantStationPoint_Cant_Station_Point" priority="299999"/>
+        <ECProperty propertyName="CantStationPoint_Station_Hidden" typeName="double" displayLabel="Station [Hidden]" category="CantStationPoint_Cant_Station_Point" priority="299998" kindOfQuantity="cifu:STATION">
```
diegoalexdiaz, line 392: This property rename is considered a major schema upgrade that is not backwards compatible (i.e. delete + add). If you just meant to hide the property, it doesn't need to be renamed. Also, if the Connector won't write to it anymore, it is a good idea to mark it as deprecated as well (look for usage of the "Deprecated" custom-attribute in other Civil Connector schemas as an example).

Domains/4-Application/Connectors/CivilInfrastructureFramework/CifRail.ecschema.xml
```diff
@@ -326,22 +388,33 @@
...
+            <ECCustomAttributes>
+                <HiddenProperty xmlns="CoreCustomAttributes.01.00.04"/>
+            </ECCustomAttributes>
+        </ECProperty>
         <ECProperty propertyName="CantStationPoint_LeftRail" typeName="double" displayLabel="Left Rail" category="CantStationPoint_Cant_Station_Point" priority="299985" kindOfQuantity="cifu:LENGTH"/>
         <ECProperty propertyName="CantStationPoint_RightRail" typeName="double" displayLabel="Right Rail" category="CantStationPoint_Cant_Station_Point" priority="299983" kindOfQuantity="cifu:LENGTH"/>
         <ECProperty propertyName="CantStationPoint_Center" typeName="double" displayLabel="Center" category="CantStationPoint_Cant_Station_Point" priority="299984" kindOfQuantity="cifu:LENGTH"/>
-        <ECProperty propertyName="CantStationPoint_Status" typeName="string" displayLabel="Fixed" category="CantStationPoint_Cant_Station_Point" priority="299984"/>
+        <ECProperty propertyName="CantStationPoint_Status" typeName="CantStationPointAspect_CantStationPoint_Status_Enum" displayLabel="Status" category="CantStationPoint_Cant_Station_Point" priority="299984"/>
```
diegoalexdiaz, line 417: This kind of change of a property (string --> enum(int)) is not supported as a minor schema update (more on it at: [link](https://www.itwinjs.org/bis/guide/schema-evolution/schema-versioning-and-generations/#minor-schema-upgrades)). 

You'd need to deprecate/hide the string-based property and create a new property (new name) based on the enum.

SachBhosale: Updated the schema as per the suggestion.

Domains/4-Application/Connectors/CivilInfrastructureFramework/CifRail.ecschema.xml
```diff
@@ -326,23 +388,82 @@
...
+        <ECProperty propertyName="CantStationPoint_TransitionType" typeName="string" displayLabel="Object Type [Deprecated]" category="CantStationPoint_Cant_Station_Point" priority="299999">
+            <ECCustomAttributes>
+                <Deprecated xmlns="CoreCustomAttributes.01.00.04">
+                    <Description>Replaced by the new CantStationPoint_TransitionType_Enum property.</Description>
+                </Deprecated>
+                <HiddenProperty xmlns="CoreCustomAttributes.01.00.04"/>
+            </ECCustomAttributes>
+        </ECProperty>
+        <ECProperty propertyName="CantStationPoint_TransitionType_Enum" typeName="CantStationPointAspect_CantStationPoint_TransitionType_Enum" displayLabel="Object Type" category="CantStationPoint_Cant_Station_Point" priority="299999"/>
+        <ECProperty propertyName="CantStationPoint_Station_Hidden" typeName="double" displayLabel="Station [Hidden]" category="CantStationPoint_Cant_Station_Point" priority="299998" kindOfQuantity="cifu:STATION">
```
diegoalexdiaz, line 400: Just want to double-check that this is what you intended:

You're adding a new property (CantStationPoint_Station_Hidden) that is hidden from the UX.
The existing property with a similar name (CantStationPoint_Station) is left intact, showing in the UX.

Seems weird to do that, but I'm probably missing more information about the motivation behind this change.

SachBhosale: CantStationPoint_Station_Hidden is just a distance along and not the formatted station value, but CantStationPoint_Station is a calculated formatted station value, which shows the same value in iModel as in dgn.
For Cant Station Point, we are getting some of the properties from PropertyServices, and I am obtaining the remaining properties from SDK. In order to keep the same values of properties that are received through PropertyServices I am using SQL query. In this query, I am using the station value as an Identifier of the Cant Station Point. The other station point may not be unique in all cases, hence I am using CantStationPoint_Station_Hidden.

diegoalexdiaz (approved):

Bordoloi123 (approved):

PR #440, opened 2025-02-07
Reflect decisions from 2/7/2025 BWG meeting.

diegoalexdiaz:
- AnnotationTextStyle not deprecated.
- Deprecate AnnotationTextStyle.Data.
- New AnnotationTextStyle.Settings property (Json-based).
- New TextAnnotationData properties (Json-based) on TextAnnotation2d and 3d.

Captured both the tweaks to deprecations as well as the new properties into a single PR rather than two as originally mentioned in the BWG meeting, in order to link them via deprecation descriptions.

These changes are meant to be included in the upcoming release of BisCore (01.00.17), which is planned to be deployed by iTwin.js-core 5.

### DIFF
--- a/Domains/0-Core/BisCore.ecschema.xml
+++ b/Domains/0-Core/BisCore.ecschema.xml
@@ -2200,7 +2200,7 @@
         </ECProperty>
         <ECProperty propertyName="Settings" typeName="string" extendedTypeName="Json" description="Settings of the style">
             <ECCustomAttributes>
-                <HiddenProperty xmlns="BisCore.01.00.16"/>
+                <HiddenProperty xmlns="CoreCustomAttributes.01.00.03"/>
             </ECCustomAttributes>
         </ECProperty>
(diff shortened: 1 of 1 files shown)


jffmarker:
My takeaway was for now, to wholesale revert Connector changes to go back to creating TextAnnotationXd elements and data as they did before. When the new JSON-based persistence is ready, Connectors can move to adopt that.

jffmarker (dismissed):

ColinKerr (approved):
This looks good to me but we need some updated documentation in remarks to cover this topic.

Also how are we tracking the format of the json stored in the TextAnnotationData property?

diegoalexdiaz:
cc @pmconne - regarding @ColinKerr 's question: "_how are we tracking the format of the json stored in the TextAnnotationData property?_"

I imagine it will be documented on the API-side of iTwin.js-core. When that happens, please let me know and we can point to it from the remarks of BisCore.

PR #455, opened 2025-03-17
DrivewayTransitions in OpenSite.01.00.05

GoldenGlaurung:
Adds the DrivewayTransition entity class with some of the properties formerly found in the DrivewayMidVertex, and a relationship class between the two.


Domains/4-Application/OpenSite/OpenSite.ecschema.xml
```diff
@@ -143,10 +143,6 @@
 
     <ECEntityClass typeName="DrivewayMidVertex" modifier="Sealed" displayLabel="Driveway Mid Vertex" description="Vertex of a DrivewayPath that is attached to two DrivewayPathEdge instances.">
         <BaseClass>PathMidVertex</BaseClass>
-
-        <ECProperty propertyName="SideTransitionOffset" typeName="double" kindOfQuantity="rru:LENGTH" category="DrivewayMidVertex_Transition" displayLabel="Transition Offset" description="The transition offset distance." />
```
karolis-zukauskas, line 147: You can't delete properties - this is a breaking change, see https://www.itwinjs.org/bis/guide/schema-evolution/schema-versioning-and-generations/#minor-schema-upgrades

karolis-zukauskas: We probably want to hide these properties and mark them as deprecated.

Domains/4-Application/OpenSite/OpenSite.ecschema.xml
```diff
@@ -433,6 +443,16 @@
...
+    <ECRelationshipClass typeName="DrivewayMidVertexOwnsTransition" strength="embedding" strengthDirection="Forward" modifier="Sealed">
+        <BaseClass>bis:ElementOwnsChildElements</BaseClass>
+        <Source multiplicity="(1..1)" roleLabel="owns" polymorphic="false">
+            <Class class="DrivewayMidVertex" />
+        </Source>
+        <Target multiplicity="(2..2)" roleLabel="is owned by" polymorphic="false">
+            <Class class="DrivewayTransition" />
+        </Target>
+    </ECRelationshipClass>
+
```
karolis-zukauskas, line 455: I don't think we want to make a change and immediately release a new Schema version - we need to test this schema first. I think normally, you make changes to the WIP (work in progress) schema, new `dev` version of this schema is published to NPM, we test it, perhaps make adjustments, and then we make a release PR.

PR #468, opened 2025-04-01
[OpenSite] Add DrivewayTransitions; deprecate old transition properties on DrivewayMidVertex

GoldenGlaurung:
Adds the DrivewayTransition entity class (owned by the DrivewayMidVertex) with offset and radius properties. These properties on the DrivewayMidVertex have been marked as deprecated.

### DIFF
--- a/Domains/4-Application/OpenSite/OpenSite.ecschema.xml
+++ b/Domains/4-Application/OpenSite/OpenSite.ecschema.xml
@@ -482,7 +482,7 @@
         </Target>
     </ECRelationshipClass>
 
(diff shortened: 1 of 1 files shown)


Domains/4-Application/OpenSite/OpenSite.ecschema.xml
```diff
@@ -447,6 +482,16 @@
         </Target>
     </ECRelationshipClass>
 
+    <ECRelationshipClass typeName="DrivewayMidVertexOwnsTransition" strength="embedding" strengthDirection="Forward" modifier="Sealed">
```
karolis-zukauskas, line 485: Perhaps `DrivewayMidVertexOwnsTransitions` (s at the end)

karolis-zukauskas (approved):

PR #483, opened 2025-04-25
Introduce Recipe3dInPlan, Represents, ComposesSubTypes and DefinitionThumbnail classes evolved by OpenSite+.

diegoalexdiaz:


simihartstein (approved):

Domains/0-Core/BisCore.ecschema.xml
```diff
@@ -877,6 +896,33 @@
...
+            <ECCustomAttributes>
+                <HiddenProperty xmlns="CoreCustomAttributes.01.00.03"/>
+            </ECCustomAttributes>
+        </ECProperty>
+        <ECProperty propertyName="ImageFormat" typeName="string" displayLabel="Image Format" description="Format of the Image captured as MIME-type.">
+            <ECCustomAttributes>
+                <HiddenProperty xmlns="CoreCustomAttributes.01.00.03"/>
+            </ECCustomAttributes>
+        </ECProperty>
+    </ECEntityClass>
```
ColinKerr, line 916: I know a thumbnail is a different use case but we should consider how this compares to how we represent textures.

https://github.com/iTwin/bis-schemas/blob/cdb0067a686d619074b856799ad725affbcf0e66/Domains/0-Core/BisCore.ecschema.xml#L2281

The referenced ImageSourceFormat enum in iTwin.js supports jpg, png and svg: https://github.com/iTwin/itwinjs-core/blob/9b71eebc6663188ae999f9c8d919f278660048ae/core/common/src/Image.ts#L105

Domains/0-Core/BisCore.remarks.md
```diff
@@ -553,3 +553,17 @@ An Entity is modeled as a `bis:RoleElement` when a set of external circumstances
 > Behavior: The system handler (C++) for `RoleElement` will only permit instances to be inserted into a `RoleModel`.
 This behavior applies to all `RoleElement` subclasses.
 `FunctionalElement` (from the `Functional` schema) is the most widely known subclass of `RoleElement`.
+
+## Relationship Classes
+
+### PhysicalTypeHasAdditionalRecipes
+
+The `RecipeDefinitionElement` already associated to a `TypeDefinition` via its _Recipe_ navigation-property shall not be also included in the associations captured by `PhysicalTypeHasAdditionalRecipes` instances.
```
ColinKerr, line 561: I think we need to include a diagram of how we intend to use this.  Maybe we can use the 3d plan recipe.

Domains/0-Core/BisCore.ecschema.xml
diegoalexdiaz: Introduced the supported formats in both cases as enums.

Domains/0-Core/BisCore.ecschema.xml
```diff
@@ -2333,6 +2338,11 @@
         <ECProperty propertyName="Data" typeName="string" extendedTypeName="Json" description="Encoded style properties" />
     </ECEntityClass>
 
+    <ECEnumeration typeName="TextureFormat" backingTypeName="int" isStrict="true" description="Supported Formats for texture data.">
```
ColinKerr, line 2341: We might have to set isStrict to false because we are updating a property that was previously just an int to be an enum.

diegoalexdiaz: Good catch. Thanks.

ColinKerr (approved):

PR #491, opened 2025-05-01
Add Curb Types to OpenSite+ Schema

dlew72:
- Adds curb type to schema with set of properties needed to calculate geometry


rbreukelaar (commented):
Looks good.

rbreukelaar (approved):

PR #499, opened 2025-05-07
Deprecate the DoGrade property on GradingArea

FelixGirard:
Kevin Jackson mentioned that this property should not be used.


rbreukelaar (commented):
ok

rbreukelaar (approved):

PR #513, opened 2025-05-15
OpenSite+ Curb Types

dlew72:
- Add navigation properties to select "smart objects" to point to a curb type to use for curb generation

image


Domains/4-Application/OpenSite/OpenSite.ecschema.xml
```diff
@@ -372,6 +372,17 @@
         <BaseClass>PathEdge</BaseClass>
     </ECEntityClass>
 
+    <ECRelationshipClass typeName="WireGeneratesCurbOfType" strength="referencing" modifier="Sealed">
+        <BaseClass>bis:ElementRefersToElements</BaseClass>
```
diegoalexdiaz, line 376: Navigation properties can't be backed by a link-table relationship (e.g. one subclassing ElementRefersToElements) but by a "foreign-key" relationship - which in this case means no base class.

I also believe EC requires each Nav-property to have a different relationship class backing it up. Thus, you likely need relationship classes such as "CurbTypeToUseInIslands" and "CurbTypeToUseInParking" rather than a single "WireGeneratesCurbOfType" relationship to back both nav-properties. Note that the two relationship classes could share a common base class if needed (e.g. for querying any subclasses in one ECSQL statement), but you might not really have such use-case.

diegoalexdiaz: It is true that the source and target classes should match the naming of the relationship. But the querying direction needs to make sense, which one can control with the "strengthDirection" attribute of a relationship class (e.g. forward vs. backward). Having said that, I recall getting confused very quickly while trying to tweak strengthDirection in both a nav-property as well as its backing relationship, so I typically take the easiest route (i.e. relationship name that matches a forward-direction querying).

If you want to give the "backward" direction a try, you may need help from @ColinKerr to get it right. Otherwise, sticking with "XXXGeneratesCurbsOfType" naming would lead to "forward-direction" relationships (easy path).

dlew72: I'll stick to the easy/simple path. Though I might switch "GeneratesCurbOfType" for "UsesCurbOfType" or "HasCurbOfType"... needs more thought

diegoalexdiaz: I recall several years ago it was decided that the usage of the verb "Has" in relationships was too vague. Thus, I think there is a schema-check on PRs that will fail if such verb is used.

dlew72: Alright, the latest version imports without errors and seems to work. Thanks for the help!

Domains/4-Application/OpenSite/OpenSite.ecschema.xml
```diff
@@ -372,6 +372,60 @@
         <BaseClass>PathEdge</BaseClass>
     </ECEntityClass>
 
+    <ECRelationshipClass typeName="DrivewayPathUsesCurbType" strength="referencing" modifier="Sealed">
+        <Source multiplicity="(0..*)" roleLabel="uses" polymorphic="true">
+            <Class class="DrivewayPath" />
+        </Source>
+        <Target multiplicity="(0..1)" roleLabel="is used by" polymorphic="true">
```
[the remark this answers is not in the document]
dlew72: I figured it'd be an easy way to:

1) support older iModels that don't have their existing smart objects linked to a curb type
2) support having "no curb" for a smart object. the editor/catalog manager thing will hopefully support "clearing" the type

Domains/4-Application/OpenSite/OpenSite.ecschema.xml
```diff
@@ -372,6 +372,60 @@
         <BaseClass>PathEdge</BaseClass>
     </ECEntityClass>
 
+    <ECRelationshipClass typeName="DrivewayPathUsesCurbType" strength="referencing" modifier="Sealed">
+        <Source multiplicity="(0..*)" roleLabel="uses" polymorphic="true">
```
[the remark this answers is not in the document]
dlew72: Done

karolis-zukauskas (approved):

PR #520, opened 2025-06-04
[OpenSite] Update Control Vertex for OpenSite schema

Emi1yZhang:
- Add `ISurfaceControlElement` mixin and `SurfaceControlOwnsControlElements` rel, which applies to all control objects and rules
  - Removed `SurfaceControlOwnsControlObjects` (they are added in last PR, not released yet so we just delete)
  - Deprecated `SurfaceControlOwnsRules`
- Combine all Control Vertex elements into `SurfaceControlVertex`
  - Delete `SurfaceSlopedControlVertex` and `SurfaceSlopedDirectionControlVertex` and migrated properties (they are added in last PR, not released yet so we just delete)
  - Add `ControlVertexType` enum for `vertexType` property of `SurfaceControlVertex`
- Add `SurfaceControlStrokeOptions` struct property for `SurfaceControlVertexRule` and `SurfaceControlLine`

### DIFF
--- a/Domains/4-Application/OpenSite/OpenSite.ecschema.xml
+++ b/Domains/4-Application/OpenSite/OpenSite.ecschema.xml
@@ -1207,7 +1207,7 @@
         </ECCustomAttributes>
     </ECEntityClass>
 
-    <ECEntityClass typeName="ISurfaceControlObject" modifier="Abstract" displayLabel="Surface Control Object" description="An interface that can be mixed-into a control vertex or control line to indicate that it can be an input for surface control.">
+    <ECEntityClass typeName="ISurfaceControlObject" modifier="Abstract" displayLabel="Surface Control Object" description="An interface that represents a spatial entity for surface control.">
         <ECCustomAttributes>
             <IsMixin xmlns="CoreCustomAttributes.01.00.03">
                 <AppliesToEntityClass>bis:SpatialLocationElement</AppliesToEntityClass>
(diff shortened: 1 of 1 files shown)


Domains/4-Application/OpenSite/OpenSite.ecschema.xml
```diff
@@ -1209,23 +1215,44 @@
         </ECCustomAttributes>
     </ECEntityClass>
 
+    <ECEntityClass typeName="ISurfaceControlElement" modifier="Abstract" displayLabel="Surface Control Element" description="An interface that can be mixed-into a control object or control rule.">
```
diegoalexdiaz, line 1218: Please expand the description of the mix-in. Right now it is not obvious what is the difference between the purpose behind `ISurfaceControlObject` and `ISurfaceControlElement`.

That confusion is leading me to currently question whether the _AppliesToEntityClass_ bit should be `bis:Element` or something higher level.

Emi1yZhang: `ISurfaceControlElement` is intended to mark any element involved in surface control, including both inputs (like control vertices and control lines) and rules that define their behavior. This is broader than ISurfaceControlObject, which is limited to input objects.
We use AppliesToEntityClass: bis:Element because the types it applies to—SurfaceControlVertex and SurfaceControlLine (from **bis:SpatialLocationElement**), and SurfaceControlRule (from **bis:InformationRecordElement**)—share bis:Element as their nearest common base class.

Domains/4-Application/OpenSite/OpenSite.ecschema.xml
```diff
@@ -1209,23 +1215,44 @@
         </ECCustomAttributes>
     </ECEntityClass>
 
+    <ECEntityClass typeName="ISurfaceControlElement" modifier="Abstract" displayLabel="Surface Control Element" description="An interface that can be mixed-into control objects and control rules to identify any entity that participates in surface control.">
+        <ECCustomAttributes>
```
diegoalexdiaz, line 1219: @Emi1yZhang , note that ISurfaceControlElement is supposed to be the base mix-in for ISurfaceControlObject. Once you introduce that change, you don't need to add ISurfaceControlElement as a mix-in for classes implementing ISurfaceControlObject (e.g. SurfaceControlVertex or Line).

PR #549, opened 2025-07-29
[OpenSite+] Added generic classification and relationship

FelixGirard:
Added a way to link classification to category to link the symbolization to the semantic.

### DIFF
--- a/Domains/4-Application/OpenSite/OpenSite.ecschema.xml
+++ b/Domains/4-Application/OpenSite/OpenSite.ecschema.xml
@@ -120,11 +120,11 @@
 
     <ECEntityClass typeName="GenericClassification" modifier="Sealed" displayLabel="Generic Classification" description="A classification that can be symbolized by a category.">
         <BaseClass>clsf:Classification</BaseClass>
-        <ECNavigationProperty propertyName="Category" relationshipName="CategorySymbolizesClassification" direction="Forward" displayLabel="Category" description="The category associated with this classification." />
+        <ECNavigationProperty propertyName="Category" relationshipName="CategorySymbolizesClassification" direction="forward" displayLabel="Category" description="The category associated with this classification." />
     </ECEntityClass>
 
     <ECRelationshipClass typeName="CategorySymbolizesClassification" strength="referencing" modifier="Abstract">
-        <Source multiplicity="(0..*)" roleLabel="symbolizes" polymorphic="false">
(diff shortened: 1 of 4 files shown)


Domains/4-Application/OpenSite/OpenSite.ecschema.xml
```diff
@@ -117,6 +118,21 @@
         <ECEnumerator value="2" name="SLOPEDIRECTION" displayLabel="Slope Direction" description="Specialized control vertex that uses slope constraints along a single direction in the spatial interpolation calculation." />
     </ECEnumeration>
 
+    <ECEntityClass typeName="GenericClassification" modifier="Sealed" displayLabel="Generic Classification" description="A classification that can be symbolized by a category.">
+        <BaseClass>clsf:Classification</BaseClass>
+        
+        <ECNavigationProperty propertyName="Category" relationshipName="CategorySymbolizesClassification" direction="Forward" displayLabel="Category" description="The category associated with this classification." />
+    </ECEntityClass>
+
+    <ECRelationshipClass typeName="CategorySymbolizesClassification" strength="referencing" modifier="Sealed">
```
diegoalexdiaz, line 127: Each nav-property requires its own separate relationship-class. Since they might need to be queried uniformly, you have the need to introduce a base-class for them (called CategorySymbolizesClassification here), in which case such base-class would need to be abstract.

FelixGirard: Why can't I use the base-class directly and need to create an abstract one?

FelixGirard: I tried adding an abstract base-class.

Domains/4-Application/OpenSite/OpenSite.ecschema.xml
```diff
@@ -118,6 +119,25 @@
         <ECEnumerator value="2" name="SLOPEDIRECTION" displayLabel="Slope Direction" description="Specialized control vertex that uses slope constraints along a single direction in the spatial interpolation calculation." />
     </ECEnumeration>
 
+    <ECRelationshipClass typeName="CategorySymbolizesClassification" strength="referencing" modifier="Abstract">
+        <Source multiplicity="(0..*)" roleLabel="symbolizes" polymorphic="false">
```
diegoalexdiaz, line 123: This relationship needs to subclass bis:ElementRefersToElements so it can be used at runtime.

diegoalexdiaz (approved):

PR #554, opened 2025-08-01
Release deprecated schema versions of the old ORDBridge.

diegoalexdiaz:


diegoalexdiaz:
@ColinKerr, note that executing 'npm run updateSchemaInventory' with the new files in this PR doesn't do anything. Is that tool aware of the "Deprecated" folders and skipping them?

PR #562, opened 2025-08-07
Add new KindOfQuantity for coordinates in AecUnits

hl662:
Adds new KindOfQuantity for coordinates in AecUnits, the FormatString is equivalent to the[ formatting config](https://github.com/iTwin/itwinjs-core/blob/261bddab0a5b42b16162ff9a57fbbe16402e89cf/core/frontend/src/quantity-formatting/QuantityFormatter.ts#L1162) used in iTwin/core-frontend's Coordinate QuantityType, both metric and imperial system.

This enables users to use a KoQ representing coordinate length, separate from KoQs used to measure length of roads or railways.


Domains/1-Common/AecUnits/AecUnits.ecschema.xml
```diff
@@ -3,7 +3,7 @@
 |  * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
 |  * See LICENSE.md in the project root for license terms and full copyright notice.
 ======================================================================================= -->
-<ECSchema schemaName="AecUnits" alias="AECU" version="01.00.04" description="This contains the unit definitions that are use across AE Schemas" displayLabel="Design Modeling Units" xmlns="http://www.bentley.com/schemas/Bentley.ECXML.3.2">
+<ECSchema schemaName="AecUnits" alias="AECU" version="01.00.05" description="This contains the unit definitions that are use across AE Schemas" displayLabel="Design Modeling Units" xmlns="http://www.bentley.com/schemas/Bentley.ECXML.3.2">
```
diegoalexdiaz, line 6: No need to increment the schema-version number. We keep WIP versions ahead of the latest released (in this case, the latest is 01.00.03).

hl662: thanks, i've reverted the change

diegoalexdiaz: That change-reverting also leads to reverting SchemaInventory.json.

hl662: just the version right? just reverted

hl662: i see, i've reverted the whole file

diegoalexdiaz (approved):

PR #565, opened 2025-08-12
Added needed Design properties to the Outfall element

Rock2000:
Took the two design fields that GravityStructure has and also created them on the Outfall. This is needed because both GravityStructure and Outfall support the NodeDesignConstraintsAspect.


BenTWhite (approved):
Looks good to me

PR #574, opened 2025-09-02
Introduce EC 3.3 JSON Schema

hl662:
Introduce EC 3.3 JSON Schema, for now, changes involve only the Format object, inside of `ecschema-item.schema.json`:
- Add bearing, ratio, azimuth to `formatType` enums
- Add `revolutionUnit`, `azimuthBase`, `azimuthBaseUnit`, `azimuthCounterClockwise` properties to `Format` for bearing and azimuth types
- Add `ratioType` for ratio type
- Add `stationBaseFactor` property for station format types 
- Updated `staionOffsetSize` to be an `integer` type, instead of `number` type. `stationBaseFactor` is using the same thing.
- Flesh out the `precision` pattern - it's now an integer, that's either from 0-12 for decimal precision, or powers of 2 for fractional precision. Utilized `oneOf` syntax to handle above


hl662:
This makes sense, I opened a PR that creates an initial jsons for ec 3.3 in draft mode: https://github.com/iTwin/bis-schemas/pull/591

And after that is merged I can update this PR to point to those drafts instead.

PR #590, opened 2025-09-29
Added SheetNumbering ECView to the SheetNumdering Schema

Ellord207:
Added ECView gathering all calculated SheeNumber Info related to Sheet.  This is intended to work with fields and provide a list of properties that should "appear to the user" as a properties of a sheet.

Questions:

1. When dealing with referenced schemas, how much does the path version matter, specifically with BisCore?  Is there a possibility of a lower patch version of a schema getting imported because of a referenced schema?
2. (Maybe more of a BWG question) Should the Sheet Id be a property on the SheetNumberInfo?  Would that make a better ECInstanceId?  Should there be some sort of relationship between a Sheet and the SheetNumberingInfo classes?

After this, we will quickly follow this PR with another looking to release this schema.

### DIFF
--- a/Domains/4-Application/DrawingProduction/SheetNumberingViews.ecschema.xml
+++ b/Domains/4-Application/DrawingProduction/SheetNumberingViews.ecschema.xml
@@ -18,165 +18,167 @@
         </SchemaLayerInfo>
     </ECCustomAttributes>
 
-    <ECEntityClass typeName="SheetIndexFolderView" modifier="Abstract" displayLabel="Group numbers" description="Group numbers for a sheet it's in a SheetIndex">
+    <ECEntityClass typeName="SheetIndexFolderView" modifier="Abstract" displayLabel="SheetIndex Folder View" description="View that returns the number of children in folder (This can include the number of children under the SheetIndex)">
         <ECCustomAttributes>
             <QueryView xmlns="ECDbMap.02.00.04">
                 <Query>
-                  SELECT
-                    [sr].ECInstanceId AS ECInstanceId,
-                    ec_classid('SheetNumbering', 'SheetIndexFolderView') [ECClassId],
-                    -- An assigned a Row Number for rows that share a parent and are ordered by their EntryPriority 
-                    --   and cast to an integer because it's a long otherwise for some reason
-                    CAST(ROW_NUMBER() OVER (PARTITION BY [sr].[Parent].[Id] ORDER BY EntryPriority) AS INTEGER) AS [GroupNumber],
-                    -- The number of SheetReferences that share a parent, again, cast to an integer
-                    CAST(COUNT([sr].[ECInstanceId]) OVER (PARTITION BY [sr].[Parent].[Id]) AS INTEGER) AS [GroupCount]
-                  FROM bis.SheetReference [sr]
+                    SELECT
+                        [sr].Parent.id AS ECInstanceId,
+                        ec_classid('SheetNumbering', 'SheetIndexFolderView') [ECClassId],
+                        -- The number of SheetReferences that share a parent, again, cast to an integer
+                        CAST(COUNT([sr].[ECInstanceId]) OVER (PARTITION BY [sr].[Parent].[Id]) AS INTEGER) AS [GroupCount]
+                    FROM bis.SheetReference [sr]
                 </Query>
             </QueryView>
         </ECCustomAttributes>
-        <ECProperty propertyName="GroupNumber" typeName="int" description="The sequential number of this sheet within its parent group, ordered by entry priority"/>
         <ECProperty propertyName="GroupCount" typeName="int" description="The total number of sheets within the same parent group as this sheet"/>
     </ECEntityClass>
-    <ECEntityClass typeName="SheetIndexView" modifier="Abstract" displayLabel="Group numbers" description="Group numbers for a sheet it's in a SheetIndex">
+    <ECEntityClass typeName="SheetIndex View" modifier="Abstract" displayLabel="SheetIndex View" description="View that returns the total number of sheets in a sheet index">
         <ECCustomAttributes>
             <QueryView xmlns="ECDbMap.02.00.04">
                 <Query>
-                  SELECT 
-                    ECInstanceId,
-                    ECClassId,
-                    PageCount
-                  FROM (
+                    SELECT 
+                        ECInstanceId,
+                        ECClassId,
+                        PageCount
+                    FROM (
+                        WITH
+                        -- General Notes:
+                        -- There are a number of times that column names are aliased arbitrarily.
+                        --  This is to avoid an issue with CTEs not respecting qualifiers.  See this issue: https://github.com/iTwin/itwinjs-core/issues/8571
+
+                        -- Recursively traverses the tree from root to leaves, building hierarchical paths.
+                        treeTraversal ([ECInstanceId], [ECClassId], [DirectParentId]) AS (
+                        -- Start with root nodes-SheetIndexes
+                        SELECT
+                            si.ECInstanceId,
+                            si.ECClassId,
+                            NULL AS [DirectParentId]
+                        FROM bis:SheetIndex si
+
+                        UNION ALL
+                            -- Recursively add children to parents, building hierarchical sort paths
+                            SELECT
+                            child.ECInstanceId,
+                            child.ECClassId,
+                            child.NodeParentId AS [DirectParentId]
+                            FROM treeTraversal parent
+                        JOIN (
+                        -- A table of all possible children joined to their parents
+                                SELECT
+                                    [sr].[ECInstanceId],
+                                    ECClassId,
+                                    [sr].[Parent].[Id] AS [NodeParentId]
+                                FROM
+                                    [bis].[SheetReference] [sr]
+
+                                UNION ALL
+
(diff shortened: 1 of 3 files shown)


Domains/4-Application/DrawingProduction/SheetNumberingViews.ecschema.xml
```diff
@@ -0,0 +1,184 @@
...
+                                    ECClassId,
+                                    [sif].[Parent].[Id] AS [NodeParentId]
+                                FROM
+                                    [bis].[SheetIndexFolder] [sif]
+                            ) child ON child.NodeParentId = parent.ECInstanceId
+                        )
+                        -- Final result: Returns only SheetReference node due to inner join
+                        SELECT
+                            sr.ECInstanceId AS [ECInstanceId],
+                            ec_classid('SheetNumbering', 'SheetIndexView') [ECClassId],
```
diegoalexdiaz, line 88: Since this view conceptually extends the persisted definition of a bis.SheetIndex, it should return its ECClassId (i.e. ec_classid('BisCore', 'SheetIndex') ) rather than the View-class' ECClassId.

Domains/4-Application/DrawingProduction/SheetNumberingViews.ecschema.xml
```diff
@@ -0,0 +1,184 @@
...
+                                [sif].[EntryPriority] AS [NodePriority],
+                                [sif].[Parent].[Id] AS [NodeParentId]
+                            FROM
+                                [bis].[SheetIndexFolder] [sif]
+                        ) child ON child.NodeParentId = parent.ECInstanceId
+                    )
+                    -- Final result: Returns only SheetReference node due to inner join
+                    SELECT
+                        sr.ECInstanceId AS [ECInstanceId],
+                        ec_classid('SheetNumbering', 'SheetReferenceView') [ECClassId],
```
diegoalexdiaz, line 167: Since this view conceptually extends the persisted definition of a bis.SheetReference, it should return its ECClassId (i.e. [sr].[ECClassId]) rather than the View-class' ECClassId.

Ellord207: Done

Domains/4-Application/DrawingProduction/SheetNumberingViews.ecschema.xml
Ellord207: Done

Domains/4-Application/DrawingProduction/SheetNumberingViews.ecschema.xml
```diff
@@ -0,0 +1,184 @@
...
+        </SchemaLayerInfo>
+    </ECCustomAttributes>
+
+    <ECEntityClass typeName="SheetIndexFolderView" modifier="Abstract" displayLabel="SheetIndex Folder View" description="View that returns the number of children in folder (This can include the number of children under the SheetIndex)">
+        <ECCustomAttributes>
+            <QueryView xmlns="ECDbMap.02.00.04">
+                <Query>
+                    SELECT
+                        [sr].Parent.id AS ECInstanceId,
+                        ec_classid('SheetNumbering', 'SheetIndexFolderView') [ECClassId],
```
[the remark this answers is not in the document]
Ellord207: Coalesces to match the container's ECClassId.

Domains/4-Application/DrawingProduction/SheetNumberingViews.ecschema.xml
```diff
@@ -0,0 +1,189 @@
...
+    <ECCustomAttributes>
+        <ProductionStatus xmlns="CoreCustomAttributes.01.00.03">
+            <SupportedUse>NotForProduction</SupportedUse>
+        </ProductionStatus>
+        <SchemaLayerInfo xmlns="BisCustomAttributes.01.00.00">
+            <Value>Application</Value>
+        </SchemaLayerInfo>
+    </ECCustomAttributes>
+
+    <ECEntityClass typeName="SheetIndexContainerView" modifier="Abstract" displayLabel="SheetIndex Container View" description="View that returns the number of children under a parent. This can be either a SheetIndex or SheetIndexFolder.">
```
diegoalexdiaz, line 21: Just wanted to double-check that you intentionally preferred the name `SheetIndexContainer`. It sounds to me like a "Container for Sheet-Indexes", which would be semantically incorrect - but if you applied a different rationale, that's fine.

Otherwise, `SheetReferenceContainer` seems more semantically correct (i.e. Container for SheetReferences).

Domains/4-Application/DrawingProduction/SheetNumberingViews.ecschema.xml
```diff
@@ -0,0 +1,189 @@
...
+        <ECCustomAttributes>
+            <QueryView xmlns="ECDbMap.02.00.04">
+                <Query>
+                    SELECT
+                        [sr].Parent.id AS ECInstanceId,
+                        COALESCE([sf].ECClassId, [si].ECClassId) AS [ECClassId],
+                        -- The number of SheetReferences that share a parent, again, cast to an integer
+                        CAST(COUNT([sr].[ECInstanceId]) OVER (PARTITION BY [sr].[Parent].[Id]) AS INTEGER) AS [GroupCount]
+                    FROM bis.SheetReference [sr] 
+                    LEFT JOIN bis.SheetIndex [si] ON si.ECInstanceId = [sr].Parent.id
```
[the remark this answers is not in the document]
Ellord207: I did this out of misplaced intent of efficiency.  After some research, I think your suggest is more efficient.

Done.

Domains/4-Application/DrawingProduction/SheetNumberingViews.ecschema.xml
Ellord207: I was following the pattern set by SheetIndexFolder.  While, in a vacuum, I agree with you, I prefer `SheetIndexContainer` given that context of `SheetIndexFolder`.

ColinKerr:
> Thanks @Ellord207. Since the ECViews technology is under evolution, we've released existing ECViews on separate "sibling" schemas that can be changed if needed, without forcing us to modify the more semantical associated schema. For example:
> 
> * [StormSewerPhysical](https://github.com/iTwin/bis-schemas/blob/master/Domains/2-DisciplinePhysical/Civil/StormSewerPhysical/StormSewerPhysical.ecschema.xml) --> [StormSewerPhysicalViews](https://github.com/iTwin/bis-schemas/blob/master/Domains/2-DisciplinePhysical/Civil/StormSewerPhysical/StormSewerPhysicalViews.ecschema.xml)
> * [BisCore](https://github.com/iTwin/bis-schemas/blob/master/Domains/0-Core/BisCore.ecschema.xml) --> [BisCoreViews](https://github.com/iTwin/bis-schemas/blob/master/Domains/0-Core/BisCoreViews.ecschema.xml)
> 
> It would be safer if you followed that pattern with yours. Also, the naming pattern we've followed with other ECView classes has included a "View" prefix.
> 
> Regarding your questions:
> 
> > 1. When dealing with referenced schemas, how much does the path version matter, specifically with BisCore? Is there a possibility of a lower patch version of a schema getting imported because of a referenced schema?
> 
> Ideally, you shall reference the minimum version of a dependency that's needed to be able to import and work with your schema. However, we don't really have any tool to calculate that yet, so it is ok that you set those versions manually. One of the schema-checks ran on PRs make sure your schema can be imported, so it will fail if the version of a dependency is not appropriate.
> 
> > 2. Should the Sheet Id be a property on the SheetNumberInfo? Would that make a better ECInstanceId?
> 
> Your ECView mixes derived properties at the SheetReference (Sheet Number, Page Number), SheetIndexFolder (Group Number, Group Page Count) and SheetIndex (Page Count) levels. While that's possible, it seems more conceptually aligned with what's persisted to split them up into three ECView classes. E.g.
> 
> * SheetReferenceView: Sheet Number, Page Number
> * SheetIndexFolderView: Group Number, Group Page Count
> * SheetIndexView: Page Count
> 
> That way, there's no confusion of what the returned ECInstanceId from each row in those ECViews mean.
> 
> > 2. Should there be some sort of relationship between a Sheet and the SheetNumberingInfo classes?
> 
> That's a known limitation of the ECView technology as it stands (cc @ColinKerr). It doesn't offer any generic metadata-way to associate a persisted class (e.g. SheetReference) with an ECView that "adds" derived properties to it (e.g. SheetReferenceView). Readers (such as your code) need to know which ECViews to join with which persisted classes. e.g.
> 
> ```
> SELECT 
>     sr.ECInstanceId,
>     sr.ECClassId,
>     sv.PageNumber
> FROM
>     SheetReference sr INNER JOIN SheetReferenceView sv ON sr.ECInstanceId = sv.ECInstanceId
> ```
> 
> However, we currently don't have any generic reader who needs to understand these associations. So, it is fine for your code (app/domain specific) to rely on its knowledge over the schemas to use these ECViews.

You can create a relationship which has a view as an endpoint: https://www.itwinjs.org/learning/ecsqlreference/views/

It is used in the ECDbMeta schema to expose custom attributes: https://github.com/iTwin/bis-schemas/blob/a16f41c0c77c988a9f169babb8a4d99979fde37a/System/ECDb/Released/ECDbMeta.04.00.03.ecschema.xml#L630

PR #591, opened 2025-10-01
Add ec3.3 draft schema

hl662:
This pull request introduces a new JSON schema definition for the EC3.3 draft representation of an `ECSchema`. The schema establishes the structure and validation rules for ECSchema objects, including required properties, references to external definitions, and constraints on additional properties.

The only changes between the 2 schema.json files in this ec3.3 draft and ec3.2 are the `$id` and `$ref` values that previously pointed to `ec/32`, updating them to `ec/33`.


rschili:
I don't think we should merge this. It will be misleading. All it does is introduce a 3.3 version husk with no context that will confuse people.

hl662:
alright, closing

PR #681, opened 2026-03-18
Introduce LayeredStructure and Pavement classes

diegoalexdiaz:


GoldenGlaurung (dismissed):

rbreukelaar (approved):
Yes, this will go a long way...

PR #694, opened 2026-04-27
Parking spaces view

FelixGirard:
- Added a new schema `OpenSiteViews` 
- Added there an ECView to calculate parking spaces counts
- Deprecate ParkingSpaces and ProblemSpaces property on LayoutParkingArea. These will be calculated using the view instead of being persisted.
- Added ProblemSpaceCount to CivilSpatial ParkingRow

### DIFF
--- a/Domains/4-Application/OpenSite/OpenSite.ecschema.xml
+++ b/Domains/4-Application/OpenSite/OpenSite.ecschema.xml
@@ -1062,8 +1062,22 @@
         <ECProperty propertyName="ParkingWidth" typeName="double" category="LayoutParkingArea_Parking" kindOfQuantity="cvu:LENGTH"  displayLabel="Width" description="Parking space width." />
         <ECProperty propertyName="ParkingDepth" typeName="double" category="LayoutParkingArea_Parking" kindOfQuantity="cvu:LENGTH"  displayLabel="Depth" description="Parking space depth." />
         <ECProperty propertyName="ParkingAngle" typeName="double" category="LayoutParkingArea_Parking" kindOfQuantity="cvu:ANGLE"  displayLabel="Angle" description="Parking space angle." />
-        <ECProperty propertyName="ParkingSpaces" typeName="int" category="LayoutParkingArea_Parking" displayLabel="Parking Spaces" description="Number of spaces." />
-        <ECProperty propertyName="ProblemSpaces" typeName="int" category="LayoutParkingArea_Parking" displayLabel="Problem Spaces" description="Number of problematic spaces." />
+        <ECProperty propertyName="ParkingSpaces" typeName="int" category="LayoutParkingArea_Parking" displayLabel="Parking Spaces" description="Number of spaces.">
+		    <ECCustomAttributes>
+                <HiddenProperty xmlns="CoreCustomAttributes.01.00.03" />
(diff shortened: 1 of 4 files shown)


Domains/4-Application/OpenSite/OpenSite.ecschema.xml
```diff
@@ -1085,6 +1099,26 @@
         <ECNavigationProperty propertyName="ParkingCurbType" relationshipName="LayoutParkingAreaUsesCurbType" direction="Forward" displayLabel="Curb Type" />
     </ECEntityClass>
 
+    <ECEntityClass typeName="ParkingInformationView" modifier="Abstract" displayLabel="Parking Information">
```
diegoalexdiaz: Since this ECView involves only Standard BIS classes, it seems to me that it could be defined in a lower-level schema so it could be helpful outside of OpenSite. That's what was done for the ECViews for StormSewerPhysical - introduced in the [StormSewerPhysicalViews schema](https://github.com/iTwin/bis-schemas/blob/master/Domains/2-DisciplinePhysical/Civil/StormSewerPhysical/StormSewerPhysicalViews.ecschema.xml).

BTW, coincidentally there is another PR introducing a "[parking space count](https://github.com/iTwin/bis-schemas/pull/635/changes#diff-bd249b2b2e7ee5b338a9ab6fe1961e5c64f8fbce034498e698f678a0d8b366c1R127)" query as part of a larger effort to provide sample ECSQL on all Standard BIS schemas. The latter uses recursion to account for any ParkingRows not directly defined under a ParkingArea. @rbreukelaar, is that never the case?

FelixGirard: I move the view to a new `OpenSiteViews` schema. Please verify this as it's the first time I'm adding a new schema.

rbreukelaar: Driveway paths can have parking rows and therefore parking spaces as well... So not all parking rows are associated to parking areas.

FelixGirard: The parking row in this case should also be contained in Parking Area @rbreukelaar 
The onside parking would generate Parking Areas that have parking rows.

rbreukelaar: Does this also apply to parking rows that the user draws? Or ones that are exploded from smarter areas?

FelixGirard: From Danny, seems like manual parking row are not really well supported at the moment, the intention is kind of it should be placed in an existing parking area....

FelixGirard: I tested and it doesn't apply to one created on their own. When created separately, the `CivilSpatial:ParkingRow` instance is driven by a `OpenSite.ParkingRowPath` and have no parent. Maybe I should support in this view separated parking rows.

FelixGirard: I updated the query to support standalone parking rows.

rbreukelaar (approved):
Ok, I think I understand this now.

PR #705, opened 2026-05-11
Delay introduction of Pavement classes

GoldenGlaurung:
Removes Pavement, PavementType, and associated relationships from the CivilPhysical schema. OpenSite will make a new class from the LayeredStructure base class in their schema.


diegoalexdiaz (dismissed):

Domains/4-Application/OpenSite/OpenSite.ecschema.xml
FelixGirard, line 15: Do you need to downgrade here?

GoldenGlaurung: This WIP OpenSite schema needs to point to the WIP CivilPhysical schema (01.00.03) for the base LayeredStructure classes and relationships

diegoalexdiaz (approved):

PR #714, opened 2026-05-27
Add new north arrow classes

scottgrahambentley:
Introduces two new classes that will be used by Drawing Production:

- `NorthArrowCallout` : A bis:AnnotationElement2d that displays a north arrow on a sheet.
- `NorthArrowCalloutDefinition` : A bis:GraphicalType2d that defines the properties of a NorthArrowCallout.

These new classes will be essential for implementing north arrow catalog functionality.

### DIFF
--- a/Domains/4-Application/DrawingProduction/DrawingProduction.ecschema.xml
+++ b/Domains/4-Application/DrawingProduction/DrawingProduction.ecschema.xml
@@ -34,8 +34,8 @@
     </ECEntityClass>
 
     <ECEntityClass typeName="NorthArrowCallout" displayLabel="North Arrow Callout" modifier="None"
-        description="A bis:DrawingGraphic that displays a north arrow on a sheet.">
-        <BaseClass>bis:DrawingGraphic</BaseClass>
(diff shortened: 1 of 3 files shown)


Ellord207:
You should probably go ahead and promote it to the [DrawingProduction](https://github.com/iTwin/bis-schemas/blob/master/Domains/4-Application/DrawingProduction/DrawingProduction.ecschema.xml) schema rather than starting it in experimental.  I don't expect there to be any changes to this between now and GA in OS+.

scottgrahambentley:
> You should probably go ahead and promote it to the [DrawingProduction](https://github.com/iTwin/bis-schemas/blob/master/Domains/4-Application/DrawingProduction/DrawingProduction.ecschema.xml) schema rather than starting it in experimental. I don't expect there to be any changes to this between now and GA in OS+.

Moved out of experimental.

Ellord207 (dismissed):
Looks like.  Might want to ask will about catalog stuff if you're not confident.

Ellord207 (approved):

PR #715, opened 2026-05-28
Fix SchemaValidation branch checkout

hl662:
## TL;DR

The Platform - Validation ADO pipeline handles PR runs and manual branch runs differently. PR runs can preserve the full source branch through `System.PullRequest.SourceBranch`, but manual and other non-PR runs fall back to `Build.SourceBranchName`, which drops everything before the final slash and breaks branch checkout for names like `scott/north-arrows`. This PR switches the fallback to the full build source ref, strips only the `refs/heads/` prefix, and fetches the requested branch before checkout so both paths use the right BisSchemas branch.

## What

| Asset | Why it exists |
|---|---|
| `tools/SchemaValidation/imodel-native-tests.yaml` | Replaces the non-PR fallback from `Build.SourceBranchName` to `Build.SourceBranch`, then strips only `refs/heads/`, so manual and other non-PR runs keep full branch names like `scott/north-arrows` instead of collapsing them to `north-arrows`. |
| `tools/SchemaValidation/imodel-native-tests.yaml` | Changes the branch switch step from a plain `git checkout %BRANCH_NAME%` to `git fetch origin "%BRANCH_NAME%"` followed by `git checkout -B "%BRANCH_NAME%" FETCH_HEAD`, so the pipeline can check out the requested branch even when the local BB-created clone does not already have that ref. |
| `tools/SchemaValidation/imodel-native-tests.yaml` | Expands the branch debug output from `git branch` to `git branch --all`, so future failures show whether the expected remote branch ref was actually fetched. |

---
*Nambot 🤖*


tools/SchemaValidation/imodel-native-tests.yaml
```diff
@@ -105,14 +111,15 @@ stages:
     - script: |
         if not "%BRANCH_NAME%" == "" (
           cd /d $(BB_SRC_ROOT)/BisSchemas
-          git checkout %BRANCH_NAME%
+          git fetch origin "%BRANCH_NAME%"
+          git checkout -B "%BRANCH_NAME%" FETCH_HEAD
```
naveedkhan8067: If BRANCH_NAME is not correct and does not exist, wouldn't it silently create a new branch and move to it?

tools/SchemaValidation/imodel-native-tests.yaml
```diff
@@ -54,13 +54,19 @@ stages:
...
         script: |
-          import sys;
-          if ( "System.PullRequest.SourceBranch" not in sys.argv[1] ):
-            branchName = sys.argv[1];
+          import sys
+
+          prBranch = sys.argv[1].strip()
+          buildBranch = sys.argv[2].strip()
+
+          if prBranch and "System.PullRequest.SourceBranch" not in prBranch:
```
naveedkhan8067, line 62: Shouldn't it be `if "System.PullRequest.SourceBranch" not in prBranch:`?

tools/SchemaValidation/imodel-native-tests.yaml
hl662: Yeah you're right, I fixed it by chaining both steps with a `|| exit /b 1`, so a missing ref to the branch name will abort the step instead of creating a new branch and moving i

tools/SchemaValidation/imodel-native-tests.yaml
hl662: No, this was intentional - The extra `prBranch and` matters when `prBranch` is empty: without it, `"System.PullRequest.SourceBranch"
not in ""` is True, so `branchRef` becomes `""` and we'd check out an empty branch name. With the guard, empty falls back to `buildBranch` too

naveedkhan8067 (approved):

PR #716, opened 2026-05-29
[OpenSiteViews] Open site parking views

FelixGirard:
Split OpenSiteParkingInformationView into one for each classes:
- LayoutParkingAreaView
- ParkingRowPathView

This should make the queries simpler and the view easier to map going forward. Having one view per class will make it easier to add more to them later.


FelixGirard:
@diegoalexdiaz  I would like your input on this and if we need a base class for the relationship like 'ElementHasView' ?

diegoalexdiaz (approved):

PR #717, opened 2026-05-31
Fix ECSchemaXML 3.2 XSD validation and add optional XSD 1.1 companion

### LINKED ISSUE #689: ECSchemaXML3.2.xsd has multiple bugs and should be validated in CI
The XSD at ECSchemaXML3.2.xsd has not been maintained since it was created (single commit - the folder rename). It currently fails to validate even BisCore.ecschema.xml. Here's what we found:

## Bugs in the XSD

**1. Typo: `Constand` in UniqueClassName selector (line 24)**
Should be `Constant`. The uniqueness constraint silently ignores Constant items.

**2. UniqueClassName selector is incomplete (lines 22-24)**
Missing `PropertyCategory`, `UnitSystem`, and `Phenomenon` - all valid schema items that should have unique typeNames.

**3. `ECSchemaReference` `name` attribute is effectively optional (line 49)**
Missing `use="required"`. A reference without a name is meaningless.

**4. `ECNavigationProperty` `relationshipName` has no type (line 556)**
Defaults to `xsd:anySimpleType`. Should be `ec:mockName`.

**5. `ec:boolean` type is case-sensitive (lines 720-728)**
Only allows lowercase `true`/`false`, but production schemas use PascalCase (`readOnly="True"`, `polymorphic="True"`). The comment mentions a style sheet conversion that doesn't exist. BisCore alone has 4 validation errors from this.

**6. `ECRelationshipDirectionType` is case-sensitive (line 175)**
Only allows `forward`/`backward`, but BisCore uses `Forward`/`Backward`.

**7. Properties after Source/Target in relationship classes cause validation failure**
The extension model for `ECRelationshipClassType` doesn't allow properties to appear after Source/Target elements, but real schemas do this (e.g. `ElementRefersToElements.MemberPriority`).

**8. `priority` attribute on properties is `xsd:string` (line 502)**
Should be `xsd:integer` - real schemas use negative values like `priority="-22"`.

**9. `PropertyCategory.priority` is `xsd:nonNegativeInteger` (line 278)**
But property-level priority supports negative values. These should be consistent.

## Proposed CI validation step

`xmllint` (libxml2) can validate ECXML 3.2 schemas against this XSD:

```bash
xmllint --noout --schema System/xsd/ECSchemaXML3.2.xsd path/to/schema.ecschema.xml
```

Once the XSD bugs are fixed, we could add a CI step that validates all non-Released `.ecschema.xml` files (or all of them) against the XSD. This would catch structural issues in schemas before they're merged.

A simple script:

```bash
#!/bin/bash
XSD="System/xsd/ECSchemaXML3.2.xsd"
ERRORS=0
for f in $(find Domains -name "*.ecschema.xml" -not -path "*/Released/*"); do
  if ! xmllint --noout --schema "$XSD" "$f" 2>/dev/null; then
    echo "FAIL: $f"
    xmllint --noout --schema "$XSD" "$f" 2>&1 | grep "validity error"
    ERRORS=$((ERRORS + 1))
  fi
done
exit $ERRORS
```

The XSD needs to be fixed first - right now it rejects valid schemas.

rschili:
Fixes: https://github.com/iTwin/bis-schemas/issues/689

## Summary

This fixes the ECSchemaXML 3.2 XSD so it better matches the ECXML that is already stored in this repository and the ECObjects readers/writers that consume it. The main XSD remains XSD 1.0 for broad tool and .NET compatibility. An optional XSD 1.1 companion is added alongside it for validators that support assertions.

## Changes

- Fix the `UniqueClassName` selector:
  - correct `Constand` to `Constant`
  - include `PropertyCategory`, `UnitSystem`, and `Phenomenon`
- Require `ECSchemaReference/@name`.
- Give `ECNavigationProperty/@relationshipName` a required `ec:mockName` type.
- Make boolean, relationship direction, relationship strength, and class modifier validation case-tolerant to match existing schemas and ECObjects parsing.
- Replace anonymous inline relationship/class attribute restrictions with named simple types so generated code has a clearer contract.
- Rework `ECRelationshipClassType` so relationship classes can contain base classes, custom attributes, properties, `Source`, and `Target` in the same content model. This allows properties that appear after `Source`/`Target`, which occurs in real schemas such as BisCore.
- Loosen relationship multiplicity from four enumerated values to the ECXML 3.2 pattern accepted by ECObjects, such as `(0..2)` and `(3..*)`.
- Use numeric priority types for property and property category priorities, while allowing the existing released-schema unsigned representations by using `xsd:long`.
- Allow the legacy `ECEnumeration/@name` alias found in existing Substation schemas, with an annotation saying new schemas should use `typeName`.
- Accept relative errors such as `.0005`.
- Add `ECSchemaXML3.2.xsd11.xsd` as an optional XSD 1.1 companion with assertions for invariants that XSD 1.0 cannot express cleanly:
  - exactly one `Source` and one `Target` on each relationship class
  - one `Class` on `Source`, at least one `Class` on `Target`
  - legacy `ECEnumeration/@name`, when present, must equal `@typeName`

## Validation

Validated the primary XSD 1.0 file with `xmllint`:

```sh
xmllint --noout System/xsd/ECSchemaXML3.2.xsd
xmllint --schema System/xsd/ECSchemaXML3.2.xsd Domains/0-Core/Released/BisCore.01.00.25.ecschema.xml --noout
```

`BisCore.01.00.25.ecschema.xml` now validates successfully. `xmllint` still prints namespace warnings for EC custom attribute namespace strings such as `BisCore.01.00.00`; those are warnings, not XSD validation errors.

I also validated all 443 ECXML 3.2 schema files found in the repository against the updated XSD 1.0 file. 435 validated successfully. The remaining 8 failures appear to be schema-file issues or tolerated legacy mistakes rather than XSD shape issues:

- unexpected `display` attributes
- unexpected `displayName` attribute
- misspelled `descriptoin` attribute
- `persistenceUnit="M(DefaultReal)"`, which is not an EC name/reference form

The XSD 1.1 companion was checked for XML well-formedness with:

```sh
xmllint --noout System/xsd/ECSchemaXML3.2.xsd11.xsd
```

Note: VS Code's XML language service may flag `xsd:override` and `xsd:assert` in the companion file because it validates schema documents using XSD 1.0 rules. The primary XSD 1.0 file is the supported broad-compatibility schema.


naveedkhan8067:
For `xmllint`, are you using the [npm package](https://www.npmjs.com/package/xmllint)?

naveedkhan8067:
Another thing we should do is to validate all future schemas against it so that they comply with it.

rschili:
> For `xmllint`, are you using the [npm package](https://www.npmjs.com/package/xmllint)?

I was using the linux default installed version, I think on windows there is also an xsd tool available. Not sure which one is the best choice, I'm not super familiar in this area to be honest

naveedkhan8067:
> > For `xmllint`, are you using the [npm package](https://www.npmjs.com/package/xmllint)?
> 
> I was using the linux default installed version, I think on windows there is also an xsd tool available. Not sure which one is the best choice, I'm not super familiar in this area to be honest

I wanted to test it locally so that we can also support it in our validations but I need to make sure the output is not different than yours (linux), I'll give it a try to this npm package. Another option that we can try is running on defualt hosted pool for linux and see how that goes.

diegoalexdiaz (approved):

naveedkhan8067 (approved):

PR #727, opened 2026-06-09
DPex Correcting Type

Ellord207:
The last release of DrawingProductionExperimental had a mistake in it.  There is a property that is suppose to store a cache of `GeometricModel.geometryGuid`, but it used `string` rather than `BeGuid`, causing the comparison of the values to always be false.


Ellord207:
@diegoalexdiaz After testing with a pre-release, I'm disappointed that the comparison between binaries doesn't seem to be helpful in ECSQL.  To have an accurate comparison of 2 binaries in a query, I have to use the following expression rather that an operator (ex. `=`, `!=`, or `<>`) directly.

`COALESCE(GuidToStr(depends.lastKnownGeometricGuid), '') <> COALESCE(GuidToStr(gm.geometryGuid), '')`

The `COALESCE` is needed since I want a situation where both cached and GeometryGuids are `NULL` to evaluate to a match.

Do you know if this is the correct way to be comparing BeGuids or is there a better way?

khanaffan:
@diegoalexdiaz  @Ellord207 

This issue stems from how `NULL` works in SQL. Unlike the traditional concept of “nil,” `NULL` represents an unknown value. As a result, comparisons like `NULL = NULL` evaluate to `FALSE`, while `NULL IS NULL` evaluates to `TRUE`. This behavior applies regardless of the underlying data type.

Some SQL dialects, including SQLite, provide an extension of the `IS` operator that allows direct comparison between columns while also handling `NULL` semantics:

```sql
col1 IS col2
col1 IS NOT col2
```

This effectively behaves like equality comparison but treats `NULL` values as comparable:

```sql
NULL IS NULL     -- TRUE
NULL IS 'foo'    -- FALSE
'foo' IS 'foo'   -- TRUE
'foo1' IS 'foo2'   -- FALSE
```

In your case, this would simplify the expression to:

```sql
WHERE depends.lastKnownGeometricGuid IS NOT gm.geometryGuid
```

ECSQL doesn’t currently support this syntax, but adding it could make comparisons like this much more straightforward and readable.

Would you be open to supporting this extension in ECSQL?

Ellord207:
@diegoalexdiaz @khanaffan

> In your case, this would simplify the expression to:
> ```sql
>  WHERE depends.lastKnownGeometricGuid IS NOT gm.geometryGuid
> ```
> ECSQL doesn’t currently support this syntax, but adding it could make comparisons like this much more straightforward and readable.
> 
> Would you be open to supporting this extension in ECSQL?

I am definitely open to this improvement. Your proposal resolves the `COALESCE`, but I'm also required to cast them both to string using `GuidToStr` regardless of if they are `NULL` or not.  Even when the properties have reportedly the same Guids, the expression `depends.lastKnownGeometricGuid = gm.geometryGuid` is still evaluate to `false`.  I assume the binary of the guids is different due to the different layers that assigned them.  Native presumably assigns `geometryGuid` and typescript assigns `lastKnownGeometricGuid`.  In the typescript, it's all just `string`.

This is actually the main concern that I have.  Sorry I didn't make this clear in my original comment.

khanaffan:
@diegoalexdiaz @Ellord207

`GeometryGuid` is stored as a `BLOB`, so comparisons should work correctly. SQLite compares blobs using `memcpy()`. I ran a test where `HexToGuid()` converts a GUID string into a blob, while `GeometryGuid` is already stored as a blob. In this setup, the record is matched successfully.

image


However, this approach fails if you compare a BLOB with a string—for example, a binary GUID versus a string GUID. This could happen if `lastKnownGeometricGuid` was previously stored as a string, or if string data was written into a binary column. SQLite will not raise an error in such cases, but comparisons will not behave as expected.

The root of the confusion is that all GUIDs in the geometry model are stored as `BeGuid` (a 16-byte binary value), whereas ECSQL and the Element API return GUIDs as strings. If you read a GUID through the Element API and store it in an instance where the property is not defined as binary with the `BeGuid` extended type, it will be persisted as a string—even if the underlying column is binary.

This PR resolves the issue by explicitly defining the column with the correct extended type:

```xml
typeName="binary" extendedTypeName="BeGuid"
```

You existing data will not work with this change. This PR correctly marking it as major version change.

Ellord207:
@diegoalexdiaz @khanaffan

>You existing data will not work with this change. This PR correctly marking it as major version change.

You're right.  I was under the miss-impression I had consume the changes using the new typing locally for testing.  After double checking from your comment, that is not the case.  Sorry for the mis-understand and thank you for setting it straight.

If you do make `IS` available, then I would appreciate it, but can work with COALESCE for now.

Thank you.

naveedkhan8067 (approved):

Alfonso-Martello (approved):

PR #736, opened 2026-06-30
Add new scale bar classes

ryanlitzke:
Introduces two new classes for DrawingProduction:

1. `ScaleBarCallout`: A bis:AnnotationElement2d that displays a scale bar on a sheet.
2. `ScaleBarCalloutDefinition`: A bis:GraphicalType2d that defines the properties of a ScaleBarCallout.

### DIFF
--- a/Domains/4-Application/DrawingProduction/DrawingProduction.ecschema.xml
+++ b/Domains/4-Application/DrawingProduction/DrawingProduction.ecschema.xml
@@ -42,4 +42,14 @@
         description="A bis:GraphicalType2d that defines the properties of a NorthArrowCallout.">
         <BaseClass>bis:GraphicalType2d</BaseClass>
     </ECEntityClass>
+    
+    <ECEntityClass typeName="ScaleBarCallout" displayLabel="Scale Bar Callout" modifier="None"
+        description="A bis:AnnotationElement2d that displays a scale bar on a sheet.">
+        <BaseClass>bis:AnnotationElement2d</BaseClass>
+    </ECEntityClass>
+
+    <ECEntityClass typeName="ScaleBarCalloutDefinition" displayLabel="Scale Bar Callout Definition" modifier="None"
+        description="A bis:GraphicalType2d that defines the properties of a ScaleBarCallout.">
+        <BaseClass>bis:GraphicalType2d</BaseClass>
+    </ECEntityClass>
 </ECSchema>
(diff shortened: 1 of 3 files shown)


Domains/4-Application/DrawingProduction/DrawingProduction.ecschema.xml
```diff
@@ -42,4 +42,14 @@
         description="A bis:GraphicalType2d that defines the properties of a NorthArrowCallout.">
         <BaseClass>bis:GraphicalType2d</BaseClass>
     </ECEntityClass>
+    
+    <ECEntityClass typeName="ScaleBarCallout" displayLabel="Scale Bar Callout" modifier="None"
+        description="A bis:AnnotationElement2d that displays a scale bar on a sheet.">
+        <BaseClass>bis:AnnotationElement2d</BaseClass>
+    </ECEntityClass>
+
+    <ECEntityClass typeName="ScaleBarCalloutDefinition" displayLabel="Scale Bar Callout Definition" modifier="None"
```
diegoalexdiaz, line 51: Just for class-naming consistency: the suffix "Definition" is used at a more abstract-level in the BisCore hierarchy. The suffix "Type" is introduced one higher level up. Having a subclass of "Type" going back to the more abstract "Definition" suffix leads to confusion.

If there is still time, consider renaming the class to "ScaleBarCalloutType" for consistency. However, not a show-stopper, so if it is too late, don't worry about it.

ryanlitzke: @diegoalexdiaz I have no issue with changing, there is still time! My one concern is that the NorthArrow uses "CalloutDefinition" and not "CalloutType", which breaks consistency between our "Callout" tools. Any insight?

diegoalexdiaz: I'm not familiar with the requirements or previous terminology in the overall topic of Callouts. Would it be acceptable to you guys rename the class but leave the displayLabel as "Callout Definition"?

ryanlitzke: @diegoalexdiaz sorry for not specifying earlier. I was referring to the way we identify the NorthArrow class here: https://github.com/iTwin/bis-schemas/blob/master/Domains/4-Application/DrawingProduction/Released/DrawingProduction.01.00.01.ecschema.xml#L36-L44

diegoalexdiaz: I see. There are Type sub-classes already released with the "Definition" suffix. In that case, let's leave it consistent within this schema.

diegoalexdiaz (approved):

PR #627, opened 2025-11-11
Release SewerHydraulicAnalysis v1.0.6

glewis-bentley:
Release SewerHydraulicAnalysis v1.0.6 with the property `FlowPath` added to the TimeOfConcentrationFlowPath element to store the geometry

PR #706, opened 2026-05-12
Release of Units 01.00.10

naveedkhan8067:
This PR contain changes to release the newer (01.00.10) of units schema.

PR #753, opened 2026-07-23
Add degree-based rotational spring constant units

### LINKED ISSUE #751: Add new ROTATIONAL_SPRING_CONSTANT unit defintions into Units schema
This request is coming from our internal structural next generation app, they are looking to use [AecUnits](https://github.com/iTwin/bis-schemas/blob/master/Domains/1-Common/AecUnits/AecUnits.ecschema.xml) schema to format their application specific properties.

They need, in particular, KN_M_PER_DEG and KPF_FT_PER_DEG, as those are the units and ways of formatting their existing user base is used to reading.

However, none of those units exist in the Units schema right now: https://github.com/iTwin/bis-schemas/blob/master/System/Units.ecschema.xml#L186-L192

Which means those won't be available for the AecUnits schema either.

We have to 
1. Add the missing units into the units schema, and release it.
2. Then, update AecUnits to consume the new units schema so they will be available.

@aruniverse @diegoalexdiaz @rschili

hl662:
## TL;DR

Structural applications cannot format rotational spring constants in the familiar kN·m/deg and kpf·ft/deg units. Adds both unit definitions and releases Units 1.0.11.

Closes #751.

## What

| Asset | Why it exists |
| --- | --- |
| `KN_M_PER_DEG` | Structural properties need familiar kN·m/deg formatting, defined from the existing `KN_M` and `ARC_DEG` units. |
| `KPF_FT_PER_DEG` | Same need for US customary, defined from the existing `KPF_FT` and `ARC_DEG` units. |
| Units 1.0.11 | Releases the two definitions so consumers can reference them. |

Advances the Units working copy to 1.0.12.

## Scope

Originally this also released AecUnits 1.0.4 with both units added to the `ROTATIONAL_SPRING_CONSTANT` presentation units. Per [review feedback](https://github.com/iTwin/bis-schemas/pull/753#issuecomment-5092363720), that has been dropped — degree-based display is not necessarily the standard presentation, so the definitions live in `Units` and the KoQ/presentation choice is left to application schemas. AecUnits is untouched by this PR.

## Companion PRs

Every Units release needs matching downstream updates ([#627](https://github.com/iTwin/imodel-native/pull/627) for 1.0.8, [#1122](https://github.com/iTwin/imodel-native/pull/1122) for 1.0.9, [#1420](https://github.com/iTwin/imodel-native/pull/1420) for 1.0.10).

- **imodel-native:** [iTwin/imodel-native#1523](https://github.com/iTwin/imodel-native/pull/1523) (draft). Covers `standardUnitCount` 529 → 531, the `SchemaWithItems` seed version, and the `BisSchemas.PartFile.xml` reference. The `SchemaSyncTest.cpp` schema hashes still need regenerating, which can only happen once this PR merges — the native build resolves Units from a sibling `BisSchemas` checkout via `SchemaInventory.json`, so until then it still sees 1.0.10. This is what the Platform - Validation failures here are pointing at.
- **itwinjs-core:** not yet opened. Precedent is [#9284](https://github.com/iTwin/itwinjs-core/pull/9284), which regenerates `core/quantity/src/generated/Units.generated.ts` and related assets.

## Also needed

Units 1.0.11 needs release admin approval/verification in `SchemaInventory.json` (currently `approved: No` / `verified: No`), which is what the released-schema validation check is reporting.


laurynas111:
I'm not sure if AecUnits is the best place to store 1/deg display unit for rotational spring constant - as that might not be the standard way to display the units.

Maybe it would be worth to add the missing units in the Units schema and then leave the KindOfQuantity for implementation in the App schemas? @diegoalexdiaz what's your thught on that?

hl662:
> Maybe it would be worth to add the missing units in the Units schema and then leave the KindOfQuantity for implementation in the App schemas?

I went ahead with this and removed the AecUnits changes, so the PR is now just the two unit definitions and the Units 1.0.11 release.

That puts it at the same three files as the previous Units releases (#706, #511): the WIP bump, the released copy, and the inventory entry.

Downstream drafts are ready for once this merges:
- [iTwin/imodel-native#1523](https://github.com/iTwin/imodel-native/issues/1523) covers `standardUnitCount` 529 to 531, the `SchemaWithItems` seed version, and the `BisSchemas.PartFile.xml` reference. The `SchemaSyncTest.cpp` hashes still need regenerating, which can only happen after this merges, since the native build resolves Units from a sibling BisSchemas checkout.

- [iTwin/itwinjs-core#9551](https://github.com/iTwin/itwinjs-core/issues/9551) has the regenerated quantity artifacts. Version bumps are held back until `@bentley/units-schema@1.0.11` publishes.

Both are intentionally failing for that reason, same as the Platform - Validation failures here in CI

SchemaInventory.json
```diff
@@ -370,6 +370,21 @@
...
       "localizations": []
+    },
+    {
+      "name": "Units",
+      "path": "System\\Released\\Units.01.00.11.ecschema.xml",
+      "released": true,
+      "version": "01.00.11",
+      "localizations": [],
+      "comment": "Add rotational spring constant degree units",
+      "verifier": "",
```
diegoalexdiaz: There is still a schema-checker that will fail if the verifier/verified fields are not set. I've questioned whether they are still needed in the age of git - @naveedkhan8067, do you know if these are still needed?

hl662: yes I was deferring updating the verifier/verified until the bis admins approves it, then I would update it myself

diegoalexdiaz:
> I'm not sure if AecUnits is the best place to store 1/deg display unit for rotational spring constant - as that might not be the standard way to display the units.
> 
> Maybe it would be worth to add the missing units in the Units schema and then leave the KindOfQuantity for implementation in the App schemas? @diegoalexdiaz what's your thught on that?

@laurynas111, please review the [KoQs in BIS](https://www.itwinjs.org/bis/guide/other-topics/kindofquantities/) article. AecUnits is meant to capture KoQs that coordinate display-formatting across the entire ecosystem. That statement is sometimes seen as too ambitious or risky for certain disciplines. Your concern seems to fall into the latter.

The Civil discipline introduced their own KoQ-oriented schema ([CivilUnits](https://github.com/iTwin/bis-schemas/blob/master/Domains/2-DisciplinePhysical/Civil/CivilUnits/CivilUnits.ecschema.xml)) for that reason. Unitized properties in OpenSite+ and Civil schemas coordinate display-formatting through CivilUnits instead of AecUnits.

Finally, keep in mind that the display-settings captured in any KoQ on those schemas are just "starting values". They don't change when users tweak settings later in the UX.

SchemaInventory.json
diegoalexdiaz: That's the chicken-egg problem of this setup. If you do that after admins approve, they will need to approve it again as your change invalidates their initial approval.

That's why these fields are always set in PRs before actual approvals in github - which raises the question whether we could get rid of them or make them optional.

naveedkhan8067: I think the entry we care about in the validator is "**approved**" not verified/verifier (may be a inventory test check for these, I have to double check otherwise we can probably make verified/verifier optional). But the reason for the **approval** flag is if we know on a later stage about a schema that its a known bad, we change the approval flag to "No" and if someone like connectors try to use that bad schema we can identify and fail before they release.

For the approvals vote reset, I think there is a setting through which we can avoid this behavour and allow once a approver has approved the vote should not reset.

laurynas111:
Thanks @diegoalexdiaz for information.
We tried similar approach initially by adding Units and KOQ to our schema. But once we did that - the schema import started failing.
Does these lines seems fine?
image

This conversation got started because that approach failed for us

diegoalexdiaz:
@laurynas111, my gut feeling is that the definition attribute doesn't support referencing of units from external schemas? Not sure. You can try referencing these new units from the WIP Units schema instead of from your side and see if that addresses it.

diegoalexdiaz (approved):

SchemaInventory.json
hl662: > we change the approval flag to "No" and if someone like connectors try to use that bad schema we can identify and fail before they release.

Do they do this programatically?

hl662: I'm gonna resolve this comment, it seems it would be a separate issue on evaluating the need for (some parts of ) the approval list

hl662:
@naveedkhan8067 can you help me understand how Platform-validation pipeline works? How the steps for getting this PR comes in? as far as I understand, I can't work on imodel-native's side to support the units until this PR comes in correct?

This is what the logs seem to mean:

```
- It finds the PR’s Units.01.00.11.ecschema.xml.
 - But imodel-native/main still names the output Units.01.00.10.ecschema.xml:
     - SYMLINK_OUT=...Units.01.00.10...
     - target: ...Units.01.00.11.ecschema.xml
 - ECObjects then tries to resolve Units.01.00.11 and cannot find it under that filename, causing the ten cascading test failures.
```

naveedkhan8067:
> @naveedkhan8067 can you help me understand how Platform-validation pipeline works? How the steps for getting this PR comes in? as far as I understand, I can't work on imodel-native's side to support the units until this PR comes in correct?
> 
> This is what the logs seem to mean:
> 
> ```
> - It finds the PR’s Units.01.00.11.ecschema.xml.
>  - But imodel-native/main still names the output Units.01.00.10.ecschema.xml:
>      - SYMLINK_OUT=...Units.01.00.10...
>      - target: ...Units.01.00.11.ecschema.xml
>  - ECObjects then tries to resolve Units.01.00.11 and cannot find it under that filename, causing the ten cascading test failures.
> ```

For this specific senario, we have to bypass this check and merge your changes. I see you already have a draft PR to fix the tests that are failing with this units schema version. If that PR is ready, I can complete this one otherwise imodel-native pipeline will be in a broken state until you fix the failing tests.

naveedkhan8067 (approved):

hl662:
> For this specific senario, we have to bypass this check and merge your changes. I see you already have a draft PR to fix the tests that are failing with this units schema version. If that PR is ready, I can complete this one otherwise imodel-native pipeline will be in a broken state until you fix the failing tests.

Yes, in that case the draft PR is ready, and I can make it active (I guess after. you complete this one?)

naveedkhan8067:
> > For this specific senario, we have to bypass this check and merge your changes. I see you already have a draft PR to fix the tests that are failing with this units schema version. If that PR is ready, I can complete this one otherwise imodel-native pipeline will be in a broken state until you fix the failing tests.
> 
> Yes, in that case the draft PR is ready, and I can make it active (I guess after. you complete this one?)

Ok I'll merge it once you are here today so that others do not get blocked in the meantime.

PR #757, opened 2026-07-28
Add KOQ for Scale Bars

ryanlitzke:
Added `KindOfQuantity` for `ScaleValue` and a `Length` property on `ScaleBarCalloutDefinition`.

All formatting of the scale value is done during runtime by building a FormatterSpec. This is because the `presentationUnits` does not currently support ratios.

### DIFF
--- a/Domains/4-Application/DrawingProduction/DrawingProduction.ecschema.xml
+++ b/Domains/4-Application/DrawingProduction/DrawingProduction.ecschema.xml
@@ -50,8 +50,7 @@
     <ECEntityClass typeName="ScaleBarCallout" displayLabel="Scale Bar Callout" modifier="None"
         description="A bis:AnnotationElement2d that displays a scale bar on a sheet.">
         <BaseClass>bis:AnnotationElement2d</BaseClass>
-        <ECProperty propertyName="Settings" typeName="string" extendedTypeName="Json" displayLabel="Settings" description="The settings for a scale bar in a versioned JSON format." />
-        <ECProperty propertyName="Scale" typeName="double" displayLabel="Scale" kindOfQuantity="SCALE_VALUE" />
+        <ECProperty propertyName="Scale" typeName="double" displayLabel="Scale" kindOfQuantity="SCALE_VALUE" description="The scale factor the scale bar was created at." />
     </ECEntityClass>
 
     <ECEntityClass typeName="ScaleBarCalloutDefinition" displayLabel="Scale Bar Callout Definition" modifier="None"
--- a/Domains/4-Application/DrawingProduction/DrawingProduction.ecschema.xml
+++ b/Domains/4-Application/DrawingProduction/DrawingProduction.ecschema.xml
@@ -9,6 +9,11 @@
     <ECSchemaReference name="BisCore" version="01.00.25" alias="bis" />
     <ECSchemaReference name="BisCustomAttributes" version="01.00.00" alias="bisCA" />
     <ECSchemaReference name="CoreCustomAttributes" version="01.00.04" alias="CoreCA" />
+    <ECSchemaReference name="Units" version="01.00.00" alias="u" />
+    <ECSchemaReference name="Formats" version="01.00.00" alias="f" />
+
+    <KindOfQuantity typeName="ScaleValue" displayLabel="Scale Value" persistenceUnit="u:COEFFICIENT" relativeError="0.0" description="A kind of quantity for scale values." />
+    <KindOfQuantity typeName="ScaleBarLength" displayLabel="Scale Bar Length" persistenceUnit="u:M" presentationUnits="f:DefaultReal(2)[u:IN];f:DefaultReal(2)[u:MM]" relativeError="0.0" description="The physical length of a scale bar on a sheet." />
     
     <ECCustomAttributes>
         <ProductionStatus xmlns="CoreCustomAttributes.01.00.03">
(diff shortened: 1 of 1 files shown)


Domains/4-Application/DrawingProduction/DrawingProduction.ecschema.xml
```diff
@@ -9,6 +9,11 @@
     <ECSchemaReference name="BisCore" version="01.00.25" alias="bis" />
     <ECSchemaReference name="BisCustomAttributes" version="01.00.00" alias="bisCA" />
     <ECSchemaReference name="CoreCustomAttributes" version="01.00.04" alias="CoreCA" />
+    <ECSchemaReference name="Units" version="01.00.00" alias="u" />
+    <ECSchemaReference name="Formats" version="01.00.00" alias="f" />
+
+    <KindOfQuantity typeName="ScaleValue" displayLabel="Scale Value" persistenceUnit="u:COEFFICIENT" relativeError="0.0" description="A kind of quantity for scale values." />
```
hl662: typeName should be all caps, and separated like `SCALE_VALUE`

Domains/4-Application/DrawingProduction/DrawingProduction.ecschema.xml
```diff
@@ -9,6 +9,11 @@
...
+    <KindOfQuantity typeName="ScaleBarLength" displayLabel="Scale Bar Length" persistenceUnit="u:M" presentationUnits="f:DefaultReal(2)[u:IN];f:DefaultReal(2)[u:MM]" relativeError="0.0" description="The physical length of a scale bar on a sheet." />
```
hl662: look at Civil units schema and etc, should typename should be aligned accordingly like `LENGTH_SCALE_BAR` and im thinking description could be clearer to denote exactly why this is different from other lengths like `LENGTH_SMALL`

Domains/4-Application/DrawingProduction/DrawingProduction.ecschema.xml
```diff
@@ -9,6 +9,11 @@
     <ECSchemaReference name="BisCore" version="01.00.25" alias="bis" />
     <ECSchemaReference name="BisCustomAttributes" version="01.00.00" alias="bisCA" />
     <ECSchemaReference name="CoreCustomAttributes" version="01.00.04" alias="CoreCA" />
+    <ECSchemaReference name="Units" version="01.00.00" alias="u" />
```
hl662: use the latest version of units, that should be 1.00.10?

Domains/4-Application/DrawingProduction/DrawingProduction.ecschema.xml
hl662: the persistence unit most likely shouldn't be represented as coefficient, it's a derivative, i'm thinking sort of like a 'slope'? https://github.com/iTwin/bis-schemas/blob/39f8005b8133521fc5443cedf0404376e7014408/System/Units.ecschema.xml#L503

ryanlitzke: @hl662 do I even need to define a persistence unit? Scale Value is really unitless. [Coefficients ](https://github.com/iTwin/bis-schemas/blob/master/System/Units.ecschema.xml#L614C1-L615C1) are the only unitless units I could find

Domains/4-Application/DrawingProduction/DrawingProduction.ecschema.xml
ryanlitzke: resolved https://github.com/iTwin/bis-schemas/pull/757/changes/c46ebf2a8dec43a592f974bbeff566b53c45502a

Domains/4-Application/DrawingProduction/DrawingProduction.ecschema.xml
ryanlitzke: resolved https://github.com/iTwin/bis-schemas/pull/757/changes/eeafb5fd56287738f84b4bdaea9c36ee9743fc87

hl662: yes, it's for the best, we have logic when configuring and editing a format that relies on the phenomenon (aka unitFamily) of the persistence unit

Domains/4-Application/DrawingProduction/DrawingProduction.ecschema.xml
ryanlitzke: good callout. I can just use `LENGTH_SMALL` in here instead of defining my own KOQ. https://github.com/iTwin/bis-schemas/pull/757/changes/b5baf2b10e371b002d49b647291400de9ef89a5e

also because we talked about it - the length of the scale bar is really a property of its definition, not it's instance. i ended up throwing it on the definition because of this, not the instance.

Domains/4-Application/DrawingProduction/DrawingProduction.ecschema.xml
ryanlitzke: @hl662 i like slope example then https://github.com/iTwin/bis-schemas/pull/757/changes/3722d60feb1e16db2e7a25758591224f8ff4d795

hl662 (approved):
approving in terms of format and koq syntax, in terms of decision to add the KoQ here over CivilUnits i'll defer to codeowners

ryanlitzke:
This PR will be closed until there is support for ratio formatting through presentation units. Future KoQ's for PAPER_LENGTH (0 precision) and SCALE_VALUE can be created as use cases for those quantities arise.

PR #761, opened 2026-08-03
rim elevations and port types to sewerphysical

BenTWhite:
1. add RimElevationAspect to be assigned to distribution structures

2. added the ability to associate Distribution Ports to Distribution Structure Types... with a new relationship


Domains/2-DisciplinePhysical/Civil/StormSewerPhysical/StormSewerPhysical.ecschema.xml
```diff
@@ -3,7 +3,7 @@
 |  * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
 |  * See LICENSE.md in the project root for license terms and full copyright notice.
 ======================================================================================= -->
-<ECSchema schemaName="StormSewerPhysical" alias="stmswrphys" version="01.00.03" description="Main schema for the physical modeling of elements used in Storm Water and Sewer Systems." xmlns="http://www.bentley.com/schemas/Bentley.ECXML.3.2">
+<ECSchema schemaName="StormSewerPhysical" alias="stmswrphys" version="01.00.04" description="Main schema for the physical modeling of elements used in Storm Water and Sewer Systems." xmlns="http://www.bentley.com/schemas/Bentley.ECXML.3.2">
```
diegoalexdiaz: No need to increment the schema-version number. It is already ahead of the last one released (01.00.02).

Domains/2-DisciplinePhysical/Civil/StormSewerPhysical/StormSewerPhysical.ecschema.xml
```diff
@@ -33,9 +33,37 @@
...
+        <Target multiplicity="(0..1)" roleLabel="owned by" polymorphic="true">
+            <Class class="RimElevationAspect"/>
+        </Target>
+    </ECRelationshipClass>
+    
     <ECEntityClass typeName="DistributionStructureType" modifier="Abstract" displayLabel="Distribution Structure Type" description="Defines a shared set of properties whose values vary per-type of DistributionStructure rather than per-instance.">
         <BaseClass>bis:PhysicalType</BaseClass>
     </ECEntityClass>
+
+    <ECRelationshipClass typeName="DistributionStructureTypeOwnsPorts" strength="embedding" description="Assign predefined ports to a distribution structure type.">
```
diegoalexdiaz: The underlying DistributionSystems schema has a [relationship](https://github.com/iTwin/bis-schemas/blob/cf53f5d1aeaa2691fc2629f74bb74624f582f720/Domains/1-Common/DistributionSystems/DistributionSystems.ecschema.xml#L80) that associates a "type of distribution element" with "a type of port". We'll need to discuss this proposal in light of that existing association.

Domains/2-DisciplinePhysical/Civil/StormSewerPhysical/StormSewerPhysical.ecschema.xml
```diff
@@ -33,9 +33,37 @@
         <BaseClass>pipphys:IPipingElement</BaseClass>
         <ECProperty propertyName="InvertElevation" typeName="double" kindOfQuantity="cvu:LENGTH" description="Elevation of the inside bottom of the Distribution Structure"/>
     </ECEntityClass>
+    
+    <ECEntityClass typeName="RimElevationAspect" modifier="None" displayLabel="RimElevationAspect">
+        <BaseClass >bis:ElementUniqueAspect</BaseClass>
+        <ECProperty propertyName="RimElevation" displayLabel="Rim Elevation" typeName="double" kindOfQuantity="cvu:LENGTH" description="Elevation at top of of distribution structure."/>
+        <ECProperty propertyName="RimToGroundElevation" displayLabel="Rim to Ground" typeName="boolean" description="If true then the rim elevation of the associated structure will automatically be set to the local ground elevation."/>
```
diegoalexdiaz: So far, Standard BIS schemas - especially on the Physical modeling perspective - have attempted to focus on semantics, leaving authoring/editing behaviors to app-schemas. I wonder if that still applies to this case.

Domains/2-DisciplinePhysical/Civil/StormSewerPhysical/StormSewerPhysical.ecschema.xml
BenTWhite: oops - right forgot how that worked - is fixed

Domains/2-DisciplinePhysical/Civil/StormSewerPhysical/StormSewerPhysical.ecschema.xml
BenTWhite: removed this for now - was going down the wrong path

Domains/2-DisciplinePhysical/Civil/StormSewerPhysical/StormSewerPhysical.ecschema.xml
BenTWhite: updated the rim aspect schema per our conversation:

image

Domains/2-DisciplinePhysical/Civil/StormSewerPhysical/StormSewerPhysical.ecschema.xml
```diff
@@ -33,9 +33,33 @@
         <BaseClass>pipphys:IPipingElement</BaseClass>
         <ECProperty propertyName="InvertElevation" typeName="double" kindOfQuantity="cvu:LENGTH" description="Elevation of the inside bottom of the Distribution Structure"/>
     </ECEntityClass>
+    
+    <ECEntityClass typeName="RimElevationAspect" modifier="Abstract" displayLabel="RimElevationAspect">
+        <BaseClass >bis:ElementUniqueAspect</BaseClass>
+    </ECEntityClass>
+
+    <ECEntityClass typeName="RimElevationSetToGroundAspect">
+        <ECProperty propertyName="RimElevation" displayLabel="Rim Elevation" readOnly="True" typeName="double" kindOfQuantity="cvu:LENGTH" description="Elevation at top of of distribution structure."/>
```
diegoalexdiaz, line 42: It is missing its base class (i.e. RimElevationAspect).

BTW, just realized the "RimElevation" can be defined in the base class (e.g. as editable) and "overriden" in the "SetToGround" subclass - setting it as read-only. The benefit in that case is that the "property definition" on the three classes is the same - no need to rely on the instance-query syntax to read it across them.

Here I'm assuming the "readOnly" attribute is part of the subset that can be overriden in a subclass - along with display-label and description. cc @rschili

BenTWhite: gotcha - wouldn't have come up with that intuitively ...  but added the missing base classes ...

PR #764, opened 2026-08-06
[OpenSite] Event Points

dlew72:
### Summary
An `EventPoint` is a rule-driven point along a `Wire` whose location is resolved by the application (TypeScript) layer, subdividing the affected wire side into span edges. The authored input is an *anchor rule* carried on a swappable aspect; the concrete aspect class determines the anchor mode (Distance / Percent / Ray / Closest).

### Design at a glance

image

Key ideas:
- **`EventPoint` *is a* `Vertex`.** Spans are ordinary `Edge` instances that reference it through the existing `EdgeStartsAtVertex` / `EdgeEndsAtVertex` pipeline.
- **Anchor mode = concrete aspect class.** Exactly one `ParametricAnchorAspect` is owned per `EventPoint`. There is no separate mode enum/discriminator.
- **Mode can change without churn.** Swapping the aspect leaves the `EventPoint`'s ElementId intact, so span edge references stay valid.
- **`EventPoint.Location`** (inherited from `Vertex`) is resolved and written by the application, not authored directly.

### Relationships

**Ownership (embedding):**

image

**Anchor references (referencing):**

image

`StartVertex` (all modes) and `EndVertex` (directed modes only) reference a polymorphic `Vertex`, so an anchor can target another `EventPoint`. For Ray/Closest, the vertex nav prop takes precedence and the positional `*Point` fallback is used only when the vertex is unset.

### Anchor modes

| Aspect (mode) | Base | StartVertex | EndVertex | Own properties | Direction source |
|---|---|---|---|---|---|
| `DistanceAnchorAspect` | `ParametricAnchorAspect` | required | — | `Distance` (signed) | sign of `Distance` |
| `ClosestAnchorAspect` | `ParametricAnchorAspect` | or `StartPoint` | — | `StartPoint` | n/a (nearest point) |
| `PercentAnchorAspect` | `DirectedAnchorAspect` | required | required | `Percent` (0..1) | anchor order Start→End |
| `RayAnchorAspect` | `DirectedAnchorAspect` | or `StartPoint` | or `EndPoint` | `StartPoint`, `EndPoint` | ray Start→End onto wire |

### What's added
- **Entity classes:** 
  - `EventPoint` (Sealed, `Vertex`)
  - `ParametricAnchorAspect` (Abstract)
  - `DirectedAnchorAspect` (Abstract)
  - `DistanceAnchorAspect` (Concrete)
  - `ClosestAnchorAspect` (Concrete)
  - `PercentAnchorAspect` (Concrete)
  - `RayAnchorAspect` (Concrete)
- **Relationships:** 
  - `EventPointOwnsParametricAnchorAspect` (embedding, 1→0..1)
  - `EventPointAnchorStartsAtVertex`  (referencing, →`Vertex` 0..1)
  - `EventPointAnchorEndsAtVertex` (referencing, →`Vertex` 0..1)
  - `WireOwnsEventPoints` (embedding, `Wire` 1→0..* `EventPoint`)
  - `EdgeOwnsSpanEdges` (embedding, `Edge`→`Edge`, for application-generated spans)
- **Property categories:**
  - `EventPoint_Anchor`
  - `DistanceAnchorAspect_Distance`
  - `PercentAnchorAspect_Percent`
  - `RayAnchorAspect_Ray`
  - `ClosestAnchorAspect_Closest`

### Consumer notes
- The application (TypeScript) resolves each `EventPoint`'s location from its anchor rule and writes `EventPoint.Location`. SITEOPS is not involved in or aware of EventPoints (other than seeing them as full fledged Vertices)
- Determine and enforce anchor mode by **which aspect is attached**
- Distance and Percent require real vertex anchors (no positional fallback). Ray and Closest accept a `Vertex` **or** a free `Point3d`; write `*Point` only when the matching vertex is unset.
- `EventPoint`s are always parented to the `Wire`

### Legend

image

### DIFF
--- a/Domains/4-Application/OpenSite/OpenSite.ecschema.xml
+++ b/Domains/4-Application/OpenSite/OpenSite.ecschema.xml
@@ -509,7 +509,7 @@
     <ECEntityClass typeName="PercentAnchorAspect" modifier="Sealed" displayLabel="Percent Anchor" description="Percent mode: a fractional position from StartVertex toward EndVertex.">
         <BaseClass>DirectedAnchorAspect</BaseClass>
 
-        <ECProperty propertyName="Percent" minimumValue="0" maximumValue="1" typeName="double" category="PercentAnchorAspect_Percent" displayLabel="Percent" description="Fractional position (0..1) along the wire from StartVertex toward EndVertex; direction comes from anchor order (authored input)." />
+        <ECProperty propertyName="Percent" minimumValue="0" maximumValue="1" kindOfQuantity="cvu:PERCENTAGE" typeName="double" category="PercentAnchorAspect_Percent" displayLabel="Percent" description="Fractional position (0..1) along the wire from StartVertex toward EndVertex; direction comes from anchor order (authored input)." />
     </ECEntityClass>
 
     <ECEntityClass typeName="RayAnchorAspect" modifier="Sealed" displayLabel="Ray Anchor" description="Ray mode: a ray from Start to End projected onto the wire.">
--- a/Domains/4-Application/OpenSite/OpenSite.ecschema.xml
+++ b/Domains/4-Application/OpenSite/OpenSite.ecschema.xml
@@ -77,6 +77,12 @@
 
     <PropertyCategory typeName="SurfaceControl" displayLabel="Surface Control" priority="0" />
 
+    <PropertyCategory typeName="EventPoint_Anchor" displayLabel="Event Point" priority="0" />
+    <PropertyCategory typeName="DistanceAnchorAspect_Distance" displayLabel="Distance" priority="0" />
+    <PropertyCategory typeName="PercentAnchorAspect_Percent" displayLabel="Percent" priority="0" />
+    <PropertyCategory typeName="RayAnchorAspect_Ray" displayLabel="Ray" priority="0" />
+    <PropertyCategory typeName="ClosestAnchorAspect_Closest" displayLabel="Closest" priority="0" />
+
     <ECEnumeration typeName="IslandType" backingTypeName="int" isStrict="true" displayLabel="Island Type" description="Describes the parking island construction type.">
         <ECEnumerator value="0" name="UNSET" displayLabel="Not Set" />
         <ECEnumerator value="1" name="LANDSCAPED" displayLabel="Landscaped" />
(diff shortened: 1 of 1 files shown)


Domains/4-Application/OpenSite/OpenSite.ecschema.xml
```diff
@@ -465,6 +471,75 @@
...
+    <ECEntityClass typeName="ClosestAnchorAspect" modifier="Sealed" displayLabel="Closest Anchor" description="Closest mode: the point on the wire closest to the anchor.">
+        <BaseClass>ParametricAnchorAspect</BaseClass>
+
+        <ECProperty propertyName="StartPoint" typeName="Point3d" category="ClosestAnchorAspect_Closest" displayLabel="Anchor Point" description="World-space anchor whose nearest point on the wire positions the EventPoint; read only when StartVertex is unset (authored input)." />
+    </ECEntityClass>
+
+    <ECEntityClass typeName="PercentAnchorAspect" modifier="Sealed" displayLabel="Percent Anchor" description="Percent mode: a fractional position from StartVertex toward EndVertex.">
+        <BaseClass>DirectedAnchorAspect</BaseClass>
+
+        <ECProperty propertyName="Percent" minimumValue="0" maximumValue="1" typeName="double" category="PercentAnchorAspect_Percent" displayLabel="Percent" description="Fractional position (0..1) along the wire from StartVertex toward EndVertex; direction comes from anchor order (authored input)." />
```
diegoalexdiaz, line 505: This unitized property is missing its KoQ. BIS wants to know more semantics over ratio-units than just "percent" - explaining why there are phenomena like SLOPE, MASS_RATIO, PROBABILITY or PERCENTAGE (fallback).

At first glance, it seemed to me that the appropriate phenomenon for this property would be LENGTH_RATIO... however:

- LENGTH_RATIO currently doesn't have any units.
- Neither VOLUME_RATIO nor MASS_RATIO offer "percentage" as a unit - as opposed to m3/m3 or Kg/Kg - which raises the question whether we should add it at all as a new unit to LENGTH_RATIO.

Thus, it seems the PERCENTAGE phenomenon might be the correct one to use here.

@hl662, @rschili

diegoalexdiaz (approved):

PR #793, opened 2026-09-08
Add new FormatSetDefinitionElement class  to BisCore

SheetalNair:
Introduce a bis class to persist formatSet in an iModel.

Refer https://github.com/iTwin/drawing-production/issues/4337 for more details

### DIFF
--- a/Domains/0-Core/BisCore.ecschema.xml
+++ b/Domains/0-Core/BisCore.ecschema.xml
@@ -2471,6 +2471,11 @@
         </Target>
     </ECRelationshipClass> 
 
+    <ECEntityClass typeName="FormatSetDefinitionElement" modifier="Sealed" displayLabel="Format Set Definition" description="A subclass of a bis:DefinitionElement that defines a formatSet.">
+        <BaseClass>DefinitionElement</BaseClass>
+        <ECProperty propertyName="Settings" typeName="string" extendedTypeName="Json" displayLabel="Settings" description="The settings for the format set in a versioned JSON format." />
+    </ECEntityClass>
+
     <ECEntityClass typeName="LineStyle" modifier="Sealed" displayLabel="Line Style">
         <BaseClass>DefinitionElement</BaseClass>
         <ECCustomAttributes>


Domains/0-Core/BisCore.ecschema.xml
```diff
@@ -2471,6 +2471,11 @@
         </Target>
     </ECRelationshipClass> 
 
+    <ECEntityClass typeName="FormatSetDefinitionElement" modifier="Sealed" displayLabel="Format Set Definition" description="A subclass of a bis:DefinitionElement that defines a formatSet.">
```
diegoalexdiaz: Following [recommendations](https://www.itwinjs.org/bis/guide/naming-guidelines/special-terms/#element) regarding the naming of non-abstract Element classes, please drop the "Element" suffix from the proposed class name (i.e. FormatSetDefinition). 

Additionally, following the last recommendation from the [general section](https://www.itwinjs.org/bis/guide/naming-guidelines/rules-and-recommendations/#general) - favor instances over definitions, maybe just "FormatSet" is sufficient since there's no "instance" equivalent to distinguish it from.

SheetalNair: updated classname to follow naming conventions

Domains/0-Core/BisCore.ecschema.xml
```diff
@@ -2471,6 +2471,11 @@
         </Target>
     </ECRelationshipClass> 
 
+    <ECEntityClass typeName="FormatSet" modifier="Sealed" displayLabel="Format Set" description="A subclass of a bis:DefinitionElement that defines a formatSet.">
```
diegoalexdiaz: In general, the term being described shouldn't appear in its description. How about:

A bis:DefinitionElement that captures the settings of a set of formats in need to be maintained in BIS Repositories.

Then, we can add more context in the BisCore.remarks.md for this class, right before its "Relationship Classes" section:

### FormatSet

A [Format](https://www.itwinjs.org/reference/ecschema-metadata/metadata/format/) is used to control how unitized values shall be displayed in the user-interface. [FormatSets](https://www.itwinjs.org/reference/ecschema-metadata/metadata/formatset/) are typically maintained outside of BIS Repositories. However, there are use-cases that need FormatSets to be stored in BIS Repositories as `DefinitionElement`s (shared data). The latter cases shall use the `FormatSet` BIS class to capture their settings.

SheetalNair: made changes

diegoalexdiaz (approved):

